Issue #71, opened 2026-09-10
VATSIM feed poller free-runs and alarms on a single transient error, so the "feed is stale" toast fires spuriously


Vainnor:
## What happens

The **"VATSIM feed is stale"** toast fires increasingly often — most noticeably while loading
historical data — even when live traffic is only seconds old. Observed toast body:

> Feed error: error sending request for url (https://data.vatsim.net/v3/vatsim-data.json). Traffic numbers may be out of date.

Two independent things drive it:

1. **A single failed poll trips the toast immediately.** The backend poller does one `vatsim::fetch`
   per tick with no retry (`backend/src/feed/vatsim.rs:204-212`). On any error it sets
   `healthy=false` + `last_error` and moves on (`backend/src/feed/mod.rs:157-162`). The client
   treats `!healthy` as stale outright — `feedIsStale` returns true the moment `healthy` is false,
   regardless of data age (`web/src/lib/feed.ts:16`, consumed in
   `web/src/components/feed-watcher.tsx:30-38`). So one transient transport hiccup ("error sending
   request for url" is a reqwest connect/timeout error, not a 5xx) surfaces the warning even though
   the last good snapshot is ~15s old.

2. **The poller is a free-running fixed interval, not phase-locked to the source.** It ticks every
   `POLL_SECS = 15` on a `tokio::time::interval` (`backend/src/feed/mod.rs:40`, `:127-129`) with no
   relation to VATSIM's own ~15s refresh (`data.general.update_timestamp`). Our tick and the
   source's refresh drift against each other, so we routinely fetch just *before* a new publish and
   then wait a full 15s — worst-case snapshot age approaches 30s at the source before the client's
   own 30s status poll and 90s `STALE_MS` threshold even see it (`feed-watcher.tsx:13`). Under the
   extra latency/contention of a historical-data load, that margin is what disappears.

## Why

- **No resilience to a transient error:** the poll is one-shot; `healthy` is a hard boolean flipped
  by a single failure, and the client has no grace window — it maps `healthy=false` straight to a
  user-facing toast. Transport blips on `data.vatsim.net` are normal and self-heal on the next tick,
  but we alarm on the first one.
- **No phase alignment:** because the interval is anchored to process start, not to the last
  observed `update_timestamp`, we don't reliably fetch *just after* the source refreshes, so we
  carry older-than-necessary data and leave less headroom before "stale."
- **Historical-load correlation is unconfirmed** and should be pinned down: whether the spike comes
[...]

PR #105, opened 2026-09-11
fix(feed): bound the poller's fast-poll streak, add success-path telemetry

### LINKED ISSUE #92: Feed poller fast-poll has no backoff ceiling, and no success-path telemetry to tune the phase-lock buffer
## What happens
`feed::poller`'s phase-lock (#71 / PR #86) schedules the next fetch ~10s after the source's last
`update_timestamp`, falling back to a 2s fast-poll once that target has passed — including while a
fetch is failing (bounded by the existing 3-consecutive-failures debounce) **and** while fetches are
succeeding but `update_timestamp` isn't advancing. The second case has no ceiling: if VATSIM ever
serves a frozen-but-200-OK response (a CDN caching a stale snapshot, an upstream bug) for an
extended period, the poller fast-polls every 2s indefinitely — a 7.5x sustained increase in request
volume against a third-party API shared by every VATSIM client, with no backoff. The old fixed-15s
ticker never had this failure mode (constant request rate regardless of content).

Separately: `REFRESH_BUFFER_SECS = 10` targets landing the fetch ~10-12s after publish. Live-tested
against the real feed during review (not reproducible in every environment): steady-state capture
lag was **~23-25s**, not ~10-12s — VATSIM's real publish-to-retrievable latency (CDN or otherwise)
is larger than the constant assumes. Still well under the client's 90s stale threshold, so the
reported bug is fixed either way, but the constant doesn't hit its own design target.

Both would be easier to diagnose and tune with actual telemetry: the poller has zero `tracing` on
its success path today (only `tracing::warn!` on fetch failure), so there's no way to observe poll
timing / chosen delay / snapshot age in production without instrumenting it first.

## What should happen
- Cap the fast-poll fallback: after N consecutive fast-polls that return `Ok` with an unchanged
  `update_timestamp`, back off to a bounded ceiling (e.g. the normal `REFRESH_BUFFER_SECS`-based
  cadence, or an explicit max backoff) rather than continuing at 2s forever. Keep the
  failure-driven fast-poll (already correctly debounced by `MAX_CONSECUTIVE_FAILURES`) as-is.
- Add `tracing::debug!`/`info!` on the success path: computed `next_poll_delay`, the resolved
  `source_timestamp`, and snapshot age at capture — the minimum needed to actually do the
  post-deploy observation PR #86 deferred to.
- Once that telemetry exists, re-tune `REFRESH_BUFFER_SECS` against real production numbers instead
  of the ~10s estimate (my one-off live sample suggested ~20-25s, but that's a single measurement,
  not a tuned constant).

## Acceptance
- [ ] A frozen-but-successful `update_timestamp` (mocked or otherwise) can't sustain 2s-interval
      polling indefinitely — a test asserts the fast-poll streak is bounded.
- [ ] Success-path tracing exists and is visible in normal ops logs at `info`/`debug`.
- [ ] `REFRESH_BUFFER_SECS` re-tuned (or explicitly left, with the reasoning recorded) using real
      captured timing data.

Blast radius: feed subsystem only (poller scheduling), same as #71. Not the trajectory/ETA model,
not permissions, not the API contract.
Pre-existing gap in #71 / PR #86; found while QA-ing it — the core debounce fix is solid and not
blocked on this.
Relates to #71.

Vainnor:
## Summary
- `backend/src/feed/mod.rs`: a frozen-but-200-OK VATSIM response never trips the existing `MAX_CONSECUTIVE_FAILURES` debounce (the fetch itself succeeds every time), so the poller could fast-poll at 2s forever — 7.5x the normal request rate against a third-party API shared by every VATSIM client, with no ceiling.
- Added `consecutive_stale_polls` tracking: once an unchanged `update_timestamp` persists for `MAX_CONSECUTIVE_STALE_POLLS` (30, ~60s — well past the ~23-25s publish latency observed live against the real feed during #71's review) fast-polls, `next_poll_delay` backs off to `STALE_POLL_BACKOFF_SECS` (30s) instead of continuing at 2s. A one-time `tracing::warn!` marks the transition.
- Added `tracing::debug!` on every successful poll (prior delay, resolved `source_timestamp`, snapshot age at capture) — the poller had zero success-path telemetry before this.
- `REFRESH_BUFFER_SECS` left at `10`, explicitly, per the issue's own stated option — the one live sample (~23-25s) is a single measurement, not a tuned constant; recorded that reasoning in its doc comment. The new telemetry is the prerequisite for retuning it responsibly.
- **Caught in self-review before push:** `consecutive_stale_polls` wasn't reset on an outright fetch failure, so a failure occurring right after a stale-freeze backoff would keep retrying at the slow 30s cadence instead of the existing fast 2s failure-retry. Fixed (second commit) — failures always reset the stale counter.

## Test plan
- `just ci`: green (160 Rust tests — 3 existing `next_poll_delay` tests updated for the new parameter, 1 new test asserting the bounded backoff per AC1, 1 self-review fix commit re-verified).
- `cargo clippy --all-targets -- -D warnings`: clean.
- Can't live-test against the real VATSIM feed from this environment (same constraint the issue itself notes for the original ~23-25s observation) — AC1 is verified by the new unit test against the pure `next_poll_delay` function; AC2/AC3 by code review (the right fields are logged; the constant's reasoning is recorded).

Closes #92


Vainnor:
QA review: rework needed.

`consecutive_stale_polls`'s update in `poller()`'s success branch doesn't match its own comment:

```rust
// A parse failure falls back to the old timestamp too — no new information either
// way, so it counts as stale the same as an unchanged value.
consecutive_stale_polls = if parsed_ts.is_some() && parsed_ts == last_source_ts {
    consecutive_stale_polls + 1
} else {
    0
};
```

The comment says a parse failure should count as stale (increment). The code does the opposite: `parsed_ts.is_some()` is `false` on a parse failure, so the whole condition is `false` and the counter **resets to 0** — every tick, indefinitely, if `update_timestamp` keeps failing to parse.

Verified with a standalone repro of the exact expression: 40 consecutive ticks of `parsed_ts = None` (fetch `Ok`, timestamp field unparseable every time) leaves `consecutive_stale_polls` pinned at 0 the entire time — `MAX_CONSECUTIVE_STALE_POLLS` (30) is never reached, so `STALE_POLL_BACKOFF_SECS` never engages, and `next_poll_delay` (whose `target` is computed from a `last_source_ts` that also never advances, since `parsed_ts.or(last_source_ts)` keeps the old value) returns the 2s fast-poll forever.

That's the same unbounded-fast-poll failure mode #92 exists to close — just triggered by a persistently-malformed `update_timestamp` instead of a frozen-but-well-formed one. The issue's own motivating scenario ("a CDN caching a stale snapshot, an upstream bug") covers this case too; AC1 ("a frozen-but-successful `update_timestamp` ... can't sustain 2s-interval polling indefinitely") isn't fully met.

Existing tests don't catch this because they only exercise the pure `next_poll_delay` function (which is correct in isolation) — nothing covers the `consecutive_stale_polls` increment/reset logic in the caller, which is where the bug actually lives.

Fix: parse failure should increment the same as an unchanged timestamp, matching the comment's stated intent — e.g. `parsed_ts.is_none() || parsed_ts == last_source_ts` (excluding only the "first successful tick ever" case, where `last_source_ts` is `None`, from counting as stale). Suggest adding a test that exercises the counter logic itself (not just `next_poll_delay`) to close this gap for future changes.

Rest of the PR is solid: `MAX_CONSECUTIVE_STALE_POLLS`/`STALE_POLL_BACKOFF_SECS` reasoning is sound, the failure-branch reset (`consecutive_stale_polls = 0` on outright fetch error) is correct, and the new tracing is exactly what AC2 asked for.

Vainnor:
Fixed — good catch. Pushed 989c1e7:

- Extracted the condition into \`is_stale_poll(parsed_ts, last_source_ts)\` and fixed it to \`parsed_ts.is_none() || parsed_ts == last_source_ts\`, exactly matching your repro and suggested fix.
- Added 4 unit tests exercising the counter logic directly (\`unchanged_timestamp_is_stale\`, \`advanced_timestamp_is_not_stale\`, \`unparseable_timestamp_is_stale_even_with_a_prior_good_one\` — the bug case — and \`first_ever_fetch_is_not_stale\`), closing the coverage gap you flagged.

\`just ci\` green (164 Rust tests). Traced the fix through your exact repro scenario by hand to confirm: 40 ticks of \`parsed_ts = None\` after a valid timestamp now correctly reaches \`MAX_CONSECUTIVE_STALE_POLLS\` and backs off, instead of pinning at 0.

PR #106, opened 2026-09-11
feat(web): enable chart hover tooltips, add a shared Tooltip for tables

### LINKED ISSUE #93: Charts and tables have no hover tooltips (built-in DOM tooltip host disabled), starting with My Dashboard
## What we want

Hover tooltips on **any chart or table across the site**, starting with the **My Dashboard** page
(`web/src/pages/dashboard.tsx` → `web/src/features/dashboard/`). Hovering a chart point/bar/slice or
a table row/cell should surface the underlying values in a themed tooltip.

## What happens today

Chart tooltips are **deliberately disabled**, but the rendering half is already written and waiting:

- The dashboard chart widget imports the tooltip-enabled `Chart` and already defines a full
  `renderTooltipBody` (themed value box: x-label + per-series swatch/label/value) —
  `web/src/features/dashboard/chart-widget.tsx:5`, `:540-564`.
- It **intentionally does not** enable the definition-level `tooltip` option. Per the in-code note
  (`chart-widget.tsx:273-278`): enabling it needs a DOM tooltip *host* extension that core looks up
  as `extension.__chartTooltipHost`, and the alpha `@tanstack/react-charts/tooltip` build
  (`@tanstack/charts` / `@tanstack/react-charts` **^0.14.0**, `web/package.json:36-37`) doesn't
  expose or inject it — turning it on **crashes at hover** with `reading '__chartTooltipHost'`. So
  today the `/tooltip` `Chart` gives only the crosshair/focus highlight, no value box.
- Tables are `@tanstack/react-table` (`web/src/features/dashboard/table-widget.tsx`, also
  `web/src/components/admin/audit-table.tsx`, `web/src/pages/planning/events.tsx`) and have no
  cell/header tooltip affordance.

## Why this is now unblocked

The reporter is right that the library supports tooltips — the framework-neutral **DOM host**
provides them. Per the TanStack Charts "DOM Host" reference, `mountChart` from
`@tanstack/charts/dom` "owns … built-in DOM tooltips," and the definition's `tooltip` option
(default `false`) drives "built-in DOM tooltip content, placement, layering, and pinning" when the
`tooltip` extension from `@tanstack/charts/tooltip` is composed into the definition. The host also
handles the positioning detail we'd otherwise fight (it sets `position: relative` on a static
container, or a `portal` extension moves the tooltip to the top layer).

So the missing piece is wiring a real DOM tooltip **host** into how we mount charts — not writing
tooltip UI (that exists). Concretely one of:

1. **Upgrade** `@tanstack/react-charts` to a version whose React `/tooltip` build injects
   `__chartTooltipHost`, then flip on the definition `tooltip` option and drop the workaround note.
2. If the React wrapper still doesn't inject a host, mount via **`mountChart` /
   `mountChartRenderer`** with `defineChart(def, { tooltip })` so the DOM host owns the tooltip, and
   feed our existing `renderTooltipBody` content through it.

Pick the smallest path that makes hover fire `renderTooltipBody` without the crash.

## What should happen

- Enable built-in chart tooltips wherever we render charts, via a DOM tooltip host — verified first
  on **My Dashboard** (line/area/bar; pie handled separately — `PieChart`, `chart-widget.tsx:524`).
- Reuse the existing `renderTooltipBody`; keep it theme-aware (`bg-popover`/`text-foreground`) and
  correct for normalized (`%`) vs raw series.
- Extend the same treatment to **tables** where a value is truncated or needs context (cell/header
  tooltips) — dashboard `table-widget.tsx` first, then the shared table users above.
- Remove the `__chartTooltipHost` workaround comment once the host is wired.
- Keep it consistent and reusable so "any chart or table" is a shared affordance, not a per-widget
  reimplementation.

## Acceptance

- [ ] Hovering a chart on My Dashboard shows a themed value tooltip (x-label + per-series
      swatch/label/value); no `__chartTooltipHost` crash.
- [ ] Tooltip content is correct for multi-series, normalized (`%`), and single-series charts, in
      light and dark themes.
- [ ] Tables expose tooltips where values are truncated/need context (dashboard table widget at
      minimum).
- [ ] The approach is shared/reusable across chart and table widgets, not one-off.
- [ ] The intentional-disable note at `chart-widget.tsx:273-278` is removed and the definition
      `tooltip` option (or DOM-host equivalent) is live.

Blast radius: web only (dashboard chart/table widgets + shared table users). No backend, no
trajectory/ETA model, no permissions, no API contract. If it requires bumping `@tanstack/charts` /
`@tanstack/react-charts`, treat that as a `dependencies` change and re-check every existing chart.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- **Chart tooltip fix** (`web/src/features/dashboard/chart-widget.tsx`): the issue was AI-drafted and speculated either a `@tanstack/charts` version bump or an imperative `mountChart` rewrite — investigated the actual installed dist instead. The crash was simply that `defineChart({...})` never composed a `tooltip` extension into the definition; `@tanstack/charts`'s currently-pinned `^0.14.0` already exports a working DOM tooltip host at its `/tooltip` subpath. Composed it in — no dependency version change needed. `renderTooltipBody` was already fully implemented and correctly shaped; it just never fired.
- **New shared `Tooltip` primitive** in `@ois/ui` (`Tooltip`/`TooltipTrigger`/`TooltipContent`/`TooltipProvider`, thin wrappers around `@radix-ui/react-tooltip`, matching the existing `avatar.tsx`/`dropdown-menu.tsx` convention), mounted once in `main.tsx`.
- **Wired into all three table call sites** (per scope decision): `table-widget.tsx` (string cells now cap width + truncate + reveal full value on hover — nothing truncated before this), `audit-table.tsx` (replaced the Time column's plain native `title={timeAgo(...)}` with the themed tooltip), `events.tsx` (same truncate+tooltip treatment on the Name column).

## Test plan
- `just ci`: green (159 Rust tests unaffected — no backend touched; `pnpm lint`/`typecheck`/`test` all pass, 82 web tests).
- Traced `@tanstack/charts`'s `renderer.js` `resolveTooltipInput` logic directly against the composed extension to confirm it satisfies the exact shape the crash was checking for — this is the verification for "no crash," since there's no way to click-and-hover in this environment.
- **Not independently verified:** actual hover behavior, tooltip positioning, and appearance in light/dark themes in a real browser (no display available here). This one especially warrants a manual check before merge — please hover a My Dashboard chart (multi-series and single-series) and all three tables in both themes.

Closes #93


Vainnor:
QA review: passing to Testing Queue, with one open item carried forward.

**Verified:**
- `__chartTooltipHost` genuinely exists in the currently-pinned `@tanstack/charts@0.14.0` dist (`dist/tooltip.js`, `dist/renderer.js`) — confirms the PR's investigation note; no library upgrade was needed, and `pnpm-lock.yaml` bears that out (only `@radix-ui/react-tooltip` is new).
- `tooltip` is a real named export of `@tanstack/charts/tooltip` in the installed package, matching the import added to `chart-widget.tsx`.
- The pre-existing `renderTooltipBody` render prop on `<Chart>` is untouched and was already correctly wired — it just had nothing to call it before `tooltip` was composed into `defineChart()`. No duplicate/competing tooltip logic.
- `TooltipProvider` is mounted above `QueryClientProvider`/`RealtimeProvider`/`RouterProvider` in `main.tsx`, so every consumer of the new `Tooltip` primitive is inside its context.
- The three table call sites (`audit-table.tsx`, `table-widget.tsx`, `events.tsx`) all reuse the one shared `packages/ui` primitive rather than one-off implementations.
- `tsc`, full test suite (82/82), and `pnpm build` all clean.

**Not verified — same gap the PR itself flagged:** actual hover/visual behavior (chart tooltip box appears without the `__chartTooltipHost` crash; content is correct for multi-series/normalized/single-series; both themes; table truncation tooltips). Every surface this touches (My Dashboard, audit log, events) sits behind VATSIM OAuth with no dev-mode bypass, so I can't reach it live — and seeding a session into the shared dev DB to fake one is correctly out of bounds for this environment (flagged by the sandbox as a shared-resource write, even with synthetic data). This needs the manual check the PR already asked for before merge — nothing here overrides that AC.

Issue #99, opened 2026-09-11
web app ships maplibre-gl 5.24.0 with a critical zero-click XSS (CVE-2026-85061)


Vainnor:
## What happens
`pnpm audit --audit-level=high` (added in #89) flags `maplibre-gl@5.24.0` — the live `web` app's map-rendering library (`web/package.json:43`, pulled directly and via `react-map-gl` → `@vis.gl/react-maplibre`) — for GHSA-jrc7-96c5-q579 / CVE-2026-85061, CVSS 10.0 (AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:N).

## Why
`DOM.sanitize()` in maplibre-gl's `src/util/dom.ts` iterates a live `NamedNodeMap` while calling `removeAttribute()` in the same loop, which shifts subsequent attributes down an index and skips one. An attacker-supplied attribution/style string with consecutive dangerous attributes (e.g. `<details open onload="1" ontoggle="...">`) has its first attribute stripped but the second survives and executes on insertion into `innerHTML` — zero-click XSS. Any code path that renders untrusted/third-party attribution or style-JSON strings through maplibre is affected. Fixed upstream in maplibre-gl 6.4.1.

## What should happen
Bump `maplibre-gl` from `"^5"` to `>=6.4.1` in `web/package.json`. This is a major version bump (5.x → 6.x), so `react-map-gl`/`@vis.gl/react-maplibre`'s compatibility with maplibre 6 needs checking, and the web app's actual map usage (layers, style specs, any attribution/style strings sourced from user or VATSIM-feed data) should be smoke-tested after the bump, not just typechecked.

## Acceptance
- [ ] `maplibre-gl` resolves to `>=6.4.1`.
- [ ] `pnpm audit --audit-level=high` no longer flags this CVE.
- [ ] Map rendering (layers, popups, attribution) is manually verified working after the bump.
- [ ] The `ignoreGhsas`/`ignoreCves` entry for this advisory added in #89 is removed.

Blast radius: none of the trajectory/ETA model, permission three-in-sync, or OpenAPI/client contract — isolated to the `web` app's map-rendering dependency.
Pre-existing; found while adding `pnpm audit` CI in #89.

Vainnor:
Shipped in #102 (targets next).

**Files:** `web/package.json` (maplibre-gl ^5 → ^6.4.1), `pnpm-lock.yaml`.

[...]

Issue #89, opened 2026-09-11
Add supply-chain CI: Dependabot, cargo-deny/audit, image scanning


Vainnor:
## What we want

Basic supply-chain hygiene in CI. OIS currently has **no** dependency automation or vulnerability
scanning: no Dependabot config, no `cargo audit`/`cargo deny`, no image scanning.

## Why

- Cargo + npm dependencies drift and accrue advisories with nothing watching them.
- The `dependencies` label already exists in the taxonomy but nothing produces those PRs.
- The images `build-images.yml` ships to GHCR (backend, web, docs) are never scanned.

## What should happen

- **Dependabot** (`.github/dependabot.yml`) for `cargo`, `npm` (pnpm workspaces), and
  `github-actions`, labeled `dependencies`. Target the `next` branch so updates flow through the
  test server like everything else.
- **`cargo deny`** (advisories, licenses, bans) and/or `cargo audit` as a CI job; **`pnpm audit`**
  for the JS side (or rely on Dependabot alerts).
- **Image scanning** (e.g. Trivy) on the images built in `build-images.yml`, failing on
  high/critical.
- Optional hardening: pin third-party actions to commit SHAs.

## Acceptance

- [ ] Dependabot opens `dependencies`-labeled PRs for cargo, npm, and actions against `next`.
- [ ] A CI job flags a known-vulnerable/denied dependency (advisories + licenses).
- [ ] Built images are scanned; a high/critical finding fails the build (or is surfaced clearly).

Blast radius: CI/infra + dependency PRs. No app-logic change. Independent of the branch-model work
beyond choosing `next` as the Dependabot target.

🤖 Drafted by Claude Code

Vainnor:
Shipped in #101 (targets next).

**Files:** `.github/dependabot.yml` (new), `deny.toml` (new) + `deny` CI job in `.github/workflows/ci.yml`, `pnpm audit` step in ci.yml's js job, `.github/workflows/build-images.yml` (Trivy scan per image), `package.json` (pnpm.auditConfig ignore-list), `web/package.json`/`pnpm-lock.yaml` (vitest 4.1.10→4.1.11).

[...]

PR #107, opened 2026-09-11
fix(discord): switch serenity's TLS backend off the vulnerable rustls-webpki

### LINKED ISSUE #98: ois-discord pulls a vulnerable rustls-webpki via serenity 0.12, 4 open RUSTSEC advisories
## What happens
`cargo deny check advisories` (added in #89) fails: 4 RUSTSEC advisories against `rustls-webpki 0.102.8` (RUSTSEC-2026-0049, -0098, -0099, -0104 — TLS certificate-verification bugs: faulty CRL distribution-point matching, incorrect URI/wildcard name-constraint handling, a reachable panic parsing CRLs).

## Why
`ois-discord` → `serenity 0.12.5` → `tokio-tungstenite 0.21.0` → `tokio-rustls 0.25.0` → `rustls 0.22.4` → `rustls-webpki 0.102.8` (`Cargo.lock:195`). `rustls 0.22.4`'s dependency spec caps `rustls-webpki` below the patched `>=0.103.10+` releases, so `cargo update -p rustls-webpki` can't resolve it — confirmed via `cargo update -p rustls-webpki@0.102.8 --dry-run` → "Locking 0 packages to highest compatible versions". A real fix needs `serenity` bumped past `0.12` (its newer releases pull a current `rustls`/`tokio-tungstenite`); `cargo update -p serenity --dry-run` currently resolves *backward* to `0.12.4` with an even older webpki, so this needs a deliberate version bump + Cargo.toml edit, not a lockfile-only update.

## What should happen
Bump `serenity` (and its transitive `tokio-tungstenite`/`rustls` stack) to a release line without these advisories, then confirm `cargo deny check advisories` passes with no `ignore` entries needed for these IDs. `discord/src/main.rs` (762 lines) should be checked for API breakage across whatever serenity major/minor line is targeted — the Discord bot is not yet live in production per AGENTS.md, so this is lower urgency than a shipped-and-exposed TLS path, but the advisories are real (certificate-verification correctness bugs).

## Acceptance
- [ ] `serenity` (and its resolved `rustls-webpki`) no longer matches RUSTSEC-2026-0049/0098/0099/0104.
- [ ] `discord/src/main.rs` compiles and its existing behavior is unchanged (or intentionally updated) across the bump.
- [ ] The 4 RUSTSEC IDs are removed from `deny.toml`'s `advisories.ignore` list (added in #89 specifically to unblock CI on this pre-existing issue).

Blast radius: none of the trajectory/ETA model, permission three-in-sync, or OpenAPI/client contract — isolated to the `ois-discord` crate's TLS/networking dependency chain.
Pre-existing; found while adding `cargo-deny` CI in #89.

Vainnor:
## Summary
- `Cargo.toml`: `serenity`'s feature flag `rustls_backend` → `native_tls_backend`.
- **The issue's own premise was wrong** (I filed it during #89): it assumed bumping `serenity` past `0.12` would pick up a patched `rustls`/`tokio-tungstenite` stack. Checked crates.io directly — `0.12.5` (already pinned) is serenity's `max_stable_version`. There is no newer release to bump to.
- serenity exposes `native_tls_backend` as a drop-in alternative to `rustls_backend` — a different transport implementation that never touches `rustls`/`rustls-webpki` at all, so it sidesteps the 4 open RUSTSEC advisories entirely rather than waiting for a patched release.
- `ois-discord`'s other networking dependency (`ois-client`'s `reqwest`) uses a separate, already-current `rustls 0.23`/`rustls-webpki 0.103.14` chain — confirmed unaffected either way.

## Verification
- Compiles clean (`cargo check -p ois-discord`) — no TLS-related code exists anywhere in `discord/src/main.rs` (762 lines, no tests either); this is purely a transport-layer feature flag, invisible to serenity's public Client/Gateway/Model API.
- `cargo deny check advisories`: clean, **zero ignore entries needed** (vs. the 4 RUSTSEC IDs `deny.toml` currently ignores from #89 — note: this branch forked from `next` before #89 merged, so `deny.toml` doesn't exist here yet; whichever of #89/#101 or this PR merges into `next` second should drop those 4 ignore entries, same sequencing note as #99).
- Checked the licenses of every new dependency this pulls (`native-tls`, `tokio-native-tls`, `hyper-tls`, `openssl-sys`) — all MIT/Apache-2.0, already within the existing allow-list.
- `just ci`: green.

## Trade-off, recorded honestly
`native-tls` uses OpenSSL via `openssl-sys` on Linux, where `rustls` needed no system library at all. `ois-discord` has no Dockerfile and isn't in `build-images.yml`'s matrix today (matches AGENTS.md: "designed, not built"), so nothing deploys this binary yet — but whenever it does get containerized, the Linux build will need OpenSSL available (system package or `openssl-sys`'s `vendored` feature).

Closes #98


Vainnor:
QA review: pass, with one cross-PR note (not a blocker for this PR).

**Verified:**
- `rustls-webpki 0.102.8` is genuinely gone from `Cargo.lock`'s resolved tree — the only `rustls-webpki` remaining is `0.103.14` (patched), from an unrelated chain (reqwest/hyper-rustls), not serenity.
- `native_tls_backend` is a real, effective feature switch, not a silent no-op typo — confirmed by the lockfile actually restructuring serenity's transitive deps (rustls/tokio-rustls/rustls-webpki chain replaced by native-tls/tokio-native-tls).
- `cargo build -p ois-discord`, `cargo clippy -p ois-discord --all-targets`, `cargo fmt --check` all clean.
- `discord/src/main.rs` untouched, consistent with this being a pure transport-layer swap under the same serenity `0.12` API surface — AC2 satisfied without code changes.
- Cross-checked against #89's `deny.toml` (still unmerged, so not part of this diff) by overlaying it onto this branch and running `cargo deny check advisories`: all 4 RUSTSEC IDs now show `advisory-not-detected` — zero live matches — confirming the fix actually closes what those ignore entries were covering.

**Note for whoever merges #89/#101 (or a follow-up):** `deny.toml`'s ignore list for RUSTSEC-2026-0049/0098/0099/0104 (its own comment already says "Remove these once #98 lands") should be dropped once both this and #101 are in — AC3 technically isn't met by this PR alone since `deny.toml` doesn't exist on `next` yet for this PR to edit. Not a blocker: confirmed `cargo-deny` only *warns* (`advisory-not-detected`) on a stale/unused ignore entry, it doesn't fail the check, so no CI breakage either merge order. Just don't let the dead entries linger past a quick cleanup pass.

PR #129, opened 2026-09-12
feat(flow): drag-to-reorder routes and FCAs on the FCA map (per-viewer)

### LINKED ISSUE #109: FCA map: let controllers manually reorder routes and FCAs by dragging
## What we want

Manually arrange the **routes** (and the **FCAs**) on the FCA page by dragging them into a preferred
order — "the little dots on the left you can grab and drag up or down to sort how you want them."

## Why / current state

- The ROUTES panel renders routes in server/default order with no drag affordance
  (`web/src/components/map/FcaMapView.tsx:648-710`); FCAs are likewise listed in a fixed order.
- Routes have no ordering field to sort by — `MapRoute` is the `RouteBody` schema
  (`web/src/lib/route.ts:7`) and `flow.route` (migration `backend/migrations/0028_flow_route.sql`)
  has no order/position column.
- **The drag pattern already exists in this codebase**: the FCA metering list reorders flights with
  dnd-kit (`SortableContext` + `useSortable` + `arrayMove`) and persists via `useReorderFca`
  (`web/src/pages/fca/detail.tsx:6`, `:261`, `:280`) — a direct model to mirror for routes/FCAs.

## The design decision to make (per-viewer vs synced)

Raised in the report; capture and decide at triage:

> **Vainnor:** per person sorting or synced across everyone?
> **jrmtx07:** … as long as people can find what they need, personal preference is better. But also
> probably not worth a ton of extra work to make it happen.

- **Per-viewer (recommended, cheap):** order lives client-side (localStorage or user preferences);
  no schema change, no permission questions. Matches the reporter's lean and "not worth a ton of
  extra work."
- **Synced (shared):** everyone sees the same order — easier to communicate/direct against, but
  needs a new order column on `flow.route` (+ FCAs), an append-only migration, a reorder endpoint,
  and a decision on who may reorder (`flow.route.update`?). Heavier.

Pick one; recommend starting per-viewer and only going synced if a shared reference order is actually
wanted.

## What should happen

- Add a drag handle (grip) to each route row and each FCA, reorder with the existing dnd-kit
  pattern, and apply the chosen order in the panel and (for routes) any list view.
- Persist per the decision above (per-viewer store, or a synced order column + endpoint).
- Keep it out of the way of the row's existing actions (Tag/Pencil/Trash) and of route visibility
  (sibling request) — dragging must not trigger edit/delete/hide.

## Acceptance

- [ ] Routes can be dragged into a custom order in the ROUTES panel; the order persists across
      reloads.
- [ ] FCAs can be dragged into a custom order.
- [ ] Persistence matches the chosen model (per-viewer, or synced across viewers) — stated in the
      PR.
- [ ] Dragging doesn't fire the row's edit/delete/label/visibility actions.

Blast radius: web (FCA map routes panel + FCA list). If **synced** is chosen, it also touches the
API contract (new reorder endpoint → client regen), `flow.route` schema (new migration), and a
permission decision — call that out. If **per-viewer**, web-only. Mirrors the existing `useReorderFca`
dnd-kit pattern. Sibling of the route-visibility request (filed separately).

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Routes and FCAs on the FCA map rendered in a fixed server/default order with no way to arrange them into a preferred sequence. Added drag-to-reorder for both, mirroring the existing dnd-kit pattern already used for FCA metering lists (`useReorderFca` in `fca/detail.tsx`).
- **Per your decision (raised in the issue thread): per-viewer, not synced.** Order lives in `localStorage` only — `web/src/components/map/lib/order-storage.ts` (`loadOrder`/`saveOrder`, mirroring the existing `view-storage.ts` convention) + `web/src/components/map/hooks/usePersistedOrder.ts` (merges the stored order against the live id set; a new route/FCA not yet ordered is appended at the end rather than disappearing). No schema change, no new endpoint, no permission question, no client regen.
- Extracted `RouteRow`/`FcaRow` as their own components — dnd-kit's `useSortable` can't be called inside a `.map()` callback.
- Dragging is isolated to the grip handle (`attributes`/`listeners` applied only there, not the row), so it never fires the row's select/edit/delete/label/visibility actions — same isolation `fca/detail.tsx`'s existing pattern relies on.

## Test plan
- `just ci`: green (168 Rust tests unaffected — no backend touched; `pnpm lint`/`typecheck`/`test` all pass, 87 web tests).
- The pure merge logic (`applyOrder`) is unit-tested directly (`order-storage.test.ts`, 5 tests: natural order, stored order applied, new items appended, deleted items dropped, empty list) rather than only exercised through the component.
- **Not independently verified:** actual drag interaction in a real browser (no display available in this environment) — worth a quick check before merge, same caveat as recent PRs.

Closes #109


Vainnor:
QA review: rework needed for the FCA (not routes) reordering.

Dragging an FCA row while the sidebar's search text or ARTCC filter is active permanently discards the persisted order for every FCA currently hidden by that filter.

Root cause: `usePersistedOrder(key, ids)` is fed `shownIds` — the **filtered** `shown` list (`filter`/`artccFilter` applied, `FcaMapView.tsx:611-622`) — not the full live FCA id set. Its setter (`setFcaOrder`) replaces the *entire* stored `order` with whatever `arrayMove` produces from the hook's current output, which itself only ever contains the currently-visible ids when a filter is active. So a drag performed under a filter overwrites `order` with just the filtered subset, silently dropping every other FCA's previously-recorded position.

(Routes don't have this problem — `routeIds` is built straight from unfiltered `routes.data`, with no filter step feeding it.)

Verified with a standalone repro of the real `applyOrder`/`arrayMove` logic:
```
initial (unfiltered) fcaOrder: [ 'a', 'b', 'c', 'd', 'e' ]
order after unfiltered drag:   [ 'b', 'c', 'a', 'd', 'e' ]   // custom order established for all 5
fcaOrder while filtered to [a,b]: [ 'b', 'a' ]
order after drag WHILE FILTERED:  [ 'a', 'b' ]                // c, d, e silently dropped from `order`
fcaOrder after clearing filter:   [ 'a', 'b', 'c', 'd', 'e' ] // c, d, e reset to natural order — their custom position is gone
```

Filtering by ARTCC (a very natural thing to do on a national FCA map before reordering "my" FCAs) is exactly the trigger. This breaks AC1 ("the order persists across reloads") in a realistic sequence: filter → drag → clear filter → the order for anything outside that filter view is gone.

Suggested fix: feed `usePersistedOrder` the **full**, unfiltered FCA id list (mirroring how routes already do it), and apply the `filter`/`artccFilter` narrowing as a separate `.filter()` step *after* ordering — so ordering always operates over every known FCA, and the visible-subset filter never touches the stored order.

Rest of the PR is solid: the `RouteRow`/`FcaRow` extraction is clean, `order-storage.ts`'s `applyOrder` correctly reconciles new/deleted ids (well covered by its 5 tests), drag activation (`PointerSensor` + 6px distance, listeners scoped only to the grip button) correctly keeps dragging from firing edit/delete/label/select — that part of AC4 is solid. This is purely the filtered-FCA-list interaction with the order hook.

Vainnor:
Fixed — good catch, and confirmed your root-cause analysis independently (routes really don't have this bug: `routeIds` is built straight from unfiltered `routes.data`, no filter step feeds it). Pushed 9b65be5:

- Ordering now applies over the full live FCA set (`allFcaIds` → `fcaOrder` → `orderedAllFcas`), with `filter`/`artccFilter` applied as a display-only step afterward (`shown`) — exactly your suggested fix.
- `onFcaDragEnd` resolves its from/to indices against the full `fcaOrder`, not the filtered `shown` list, so a drag within a filtered view still repositions correctly relative to the drop target without touching anything outside the filter.
- Added a regression test capturing the exact invariant that broke (resolving a drag's indices against the full order never drops filtered-out items), traced through your exact repro numbers.

`just ci` green (168 Rust tests, 88 web tests including the new regression test).

PR #144, opened 2026-09-12
fix(web): bring ATC layer refetch cadence in line with traffic (15s)

### LINKED ISSUE #142: ATC logon/logoff is slow to appear on maps — ATC layer refetches at 30s (2× slower than traffic)
## What happens

When a controller logs on or off, the change takes a long time (tens of seconds, up to ~1 min) to
appear/disappear on any map's ATC layer — noticeably laggier than live traffic, which updates
promptly.

## Why

The ATC map layer is driven by `useAtc` (`web/src/lib/fca.ts:324-339`), whose **live refetch is 30s**:

```ts
// web/src/lib/fca.ts:335
refetchInterval: at == null ? 30_000 : false,
```

The live traffic/flow query in the same file refetches at **15s** (`fca.ts:291`, `:307`) — so ATC
polls at *half* the rate of traffic for no structural reason. The `/api/v1/flow/atc` endpoint is
cheap: `handlers/atc.rs::list_atc` just classifies the in-memory feed snapshot (controllers/atis) —
no DB query — so a 30s client interval is conservative.

Total observed lag is the sum of three stages; the client refetch is the largest tunable one:
1. **VATSIM publish latency** — ~15–25s for the datafeed itself to reflect the logon/logoff (external, fixed).
2. **Backend feed poll** — up to `POLL_SECS = 15` (`feed/mod.rs`) to ingest the new snapshot.
3. **Client refetch** — up to **30s** (`useAtc`) to fetch the new board.

Worst case ≈ 25 + 15 + 30 ≈ ~70s; typical ~45–60s. Stage 3 is the cheapest to halve.

## What should happen

Bring `useAtc`'s live `refetchInterval` in line with traffic — **15s** — so the ATC layer refreshes
at the same cadence as the aircraft it sits among. The endpoint cost is negligible (in-memory
classification), so this roughly halves the client-side component with no backend load concern.

Residual floor after that is stages 1–2 (VATSIM latency + the 15s poll), which are inherent; if the
backend poll's phase-lock (tuned in #92) still leaves ATC noticeably behind, that can be revisited
separately, but the 2× client interval is the clear first fix.

## Acceptance

- [ ] `useAtc` live refetch interval matches the traffic layer (15s), not 30s.
- [ ] A logon/logoff is reflected on the map within roughly the traffic-layer latency (no longer 2×
      slower than aircraft updates).
- [ ] No added backend load of note (endpoint still classifies the cached snapshot; no new queries).

Blast radius: web only (`web/src/lib/fca.ts` `useAtc`). No backend change, no API contract, no
trajectory/permissions. `/flow/atc` already serves the current snapshot; this is purely client
refetch cadence.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- `useAtc` (`web/src/lib/fca.ts`) polled at 30s while `useTraffic`/`useFcaCounts` already poll at 15s for the same map, so ATC logon/logoff lagged ~2× behind aircraft updates for no structural reason.
- The `/api/v1/flow/atc` endpoint is pure in-memory classification of the cached feed snapshot (no DB query), so halving the client interval adds no backend load.
- Updated the doc comment above `useAtc`, which previously claimed the 30s interval was deliberate due to "heavier geometry" — that didn't hold up: the ATC layer draws at most a few dozen ARTCC/TRACON polygons, while traffic already redraws far more every 15s.

## Test plan
- No visual/browser check possible in this environment; the change is a single constant swap in an existing ternary (no new logic path), so typecheck + full test suite passing is the available evidence.

Closes #142


Vainnor:
## QA review — PASS

Tested in a worktree on `fix/142/atc-refetch-cadence` (1 commit ahead of `next`, up to date, no conflicts).

**Verified claims:**
- `useTraffic`/`useFcaCounts` do poll at 15s live — the "match traffic cadence" rationale checks out (`web/src/lib/fca.ts:291,307`).
- `list_atc` (`backend/src/handlers/atc.rs`) is pure in-memory classification of the cached feed snapshot (`state.feed`, `state.tracons` — ArcSwap/RwLock reads) — no DB query, confirming "no added backend load" holds even at 2× frequency.
- No API contract change (no schema/openapi diff) — client regen not required.
- Historical path (`at != null`) unaffected — `refetchInterval` stays `false` there either way.

**Ran:**
- `just ci` (fmt-check, cargo check, 168 Rust tests, pnpm typecheck) — all green.
- Mutation test: reverted `useAtc`'s interval to `30_000` — all 88 web tests still passed, confirming no test currently pins this specific value (not a regression from this PR; the file's other intervals are similarly untested — pre-existing pattern, not introduced here).

**Non-blocking observation:** `useAtc` is also called unconditionally (`useAtc(true)`) by the always-mounted dashboard `AtcWidgetView` (`web/src/features/dashboard/atc-widget.tsx:40`, rendered via `dashboard/render.tsx`) — not just the two map layers that gate on `showAtc` (`FcaMapView.tsx:369`, facility map). This PR's framing (comment, commit message, issue body) is entirely about "the map ATC layer," but the change also silently doubles polling for that dashboard widget, independent of any map being open. Functionally harmless given the endpoint's cost, and arguably a net positive (faster widget refresh too) — flagging for awareness, not requesting rework.

No security, permission, migration, or trajectory-model concerns — route pre-existed unchanged, no state mutation, no new DB access.

**Verdict:** Acceptance criteria met. No regressions. Approved by operator.

PR #156, opened 2026-09-12
fix(map): let clicks pass through the MapLibre basemap to deck.gl

### LINKED ISSUE #154: Cannot click aircraft on any map — deck.gl pick no longer resolves (regression after map-stack bumps)
## What happens

On **every** map, clicking an aircraft no longer does anything — no selection, no callsign focus.
Regression; it worked before the recent map-stack changes.

## Why (suspected)

The click path is intact in code, so a deck.gl **pick** is no longer resolving to the aircraft layer:

- The aircraft glyphs are a `pickable: true` deck.gl `IconLayer` — `web/src/components/map/layers/aircraft.ts:34-37`.
- deck's `onClick` handler branches on the picked layer id — `web/src/components/map/TrafficMap.tsx:221-223`:
  ```ts
  if (info.layer?.id === "aircraft") {
    const id = (info.object as NormAircraft | undefined)?.id;
    if (id) onAircraftClick?.(id);
  }
  ```
- deck + the MapLibre basemap are composited in `web/src/components/map/MapCanvas.tsx` (`<DeckGL onClick={onClick}>` over `<MapLibre>`).

Since the handler and `pickable` are unchanged, the pick is coming back empty (`info.layer` undefined),
which points at the map dependency churn just merged. Prime suspects, in order:

1. **maplibre basemap fix (#139)** — before it, the basemap didn't render (black) and clicks worked;
   after it, the basemap renders *and* clicks broke. That correlation suggests the maplibre canvas now
   sits above / captures pointer events over the deck canvas (z-order / `pointer-events` / the
   interleaved-vs-overlaid compositing), so deck never sees the click.
2. **deck.gl 9.3.10 → 9.4.0** (#121/#122) — a picking/`onClick` behavior change in 9.4.
3. **react-map-gl 8.1.3** (#120) — canvas/DOM layering change in the `react-map-gl/maplibre` wrapper.

## What should happen

Clicking an aircraft selects it / focuses the callsign again. Investigation order:
1. Reproduce with the browser devtools — confirm deck's `onClick` fires at all, and inspect `info`
   (is `info.layer` empty? is the event hitting the MapLibre canvas instead of deck's?).
2. Check the deck ↔ maplibre compositing in `MapCanvas.tsx` (which canvas is on top, `pointer-events`,
   whether layers are interleaved or overlaid) — most likely lever given the #139 correlation.
3. If it's a deck 9.4 picking change, adjust accordingly; bisect by temporarily pinning deck.gl back
   to 9.3.10 and/or reverting #139 to confirm which change is responsible.

## Acceptance

- [ ] Clicking an aircraft on any map selects it / focuses its callsign (deck `onClick` resolves the
      `aircraft` layer pick again).
- [ ] Basemap still renders and hover tooltips still work (no re-regression of #137/#139).
- [ ] Root cause identified (which of the map-stack changes) and noted in the fix.

Blast radius: web/map only (`MapCanvas.tsx` compositing, `TrafficMap.tsx` click handling, map dep
stack). No backend, no API contract. Regression correlated with the recent map bumps
(deck.gl 9.4 / react-map-gl 8.1.3 / maplibre basemap fix #139). On `next` and — after the last
promotions — likely `main`.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Root cause (found by tracing actual `@deck.gl/react`/`@vis.gl/react-maplibre` rendering source, not by guessing at the recent dependency bumps): deck.gl renders the MapLibre basemap as a sibling `<div>` directly on top of its own `<canvas>` (same absolute bounds), and neither library sets `pointer-events` on it. Since MapLibre is a pure background renderer here (deck.gl owns all camera/interaction, per this file's own comment), its div/canvas was silently swallowing every click/drag before deck's own canvas — where picking and the pan/zoom controller live — ever saw it.
- The issue's own top theory (the #139 basemap fix changed compositing) is factually wrong — I read #139's actual diff, which only fixed maplibre-gl 6's worker-script URL resolution, touching zero DOM/z-order code. Checked deck.gl 9.4.0's and react-map-gl 8.1.3's real changelogs too: neither documents a picking/interaction change.
- Fix: `style={{ pointerEvents: "none" }}` on `<MapLibre>`. No native MapLibre control (`NavigationControl`, etc.) is used anywhere in this app, so this has no downside.

## Test plan
- **Could not visually confirm the click now works** — no browser/display in this environment, and this repo has no Playwright/Puppeteer or other E2E infra (checked; none present) to add one proportionately for a single CSS property. The fix is grounded in reading the actual deck.gl + react-map-gl source, not a guess, but **please do a manual click-test on a real map before merging** to confirm AC1.

Closes #154


Vainnor:
## QA review — PASS

Tested in a worktree on `fix/154/map-click-regression` (1 commit ahead of `next`, up to date, no conflicts).

**Verified the root-cause claim independently, not just trusted the description:**
- Diffed #139's actual commit (`df04790`) myself — confirmed it only adds `mapLib={mapLibPromise}` (a worker-URL loading fix) and touches zero DOM/z-order/pointer-events code. The issue's own top theory (that #139 caused this) is indeed disproven, as the PR claims.
- Read the actual `@vis.gl/react-maplibre@8.1.3` source (`dist/components/map.js`) rather than trusting the stated mechanism: confirmed the outer container div receives `style` (where `pointerEvents: "none"` now lands via `...props.style`), and `props.children` render inside a nested `mapboxgl-children` div with `CHILD_CONTAINER_STYLE = { height: '100%' }` — no pointer-events override. This is the exact mechanism described, verified against the real library code, not just plausible-sounding.
- Confirmed via grep: no `NavigationControl`/`GeolocateControl`/`ScaleControl`/`FullscreenControl`/direct `maplibregl.*` usage anywhere in the app — the "no native MapLibre control" claim holds.
- Confirmed `MapCanvas.tsx` is the *only* place `<MapLibre>` is mounted — this fix covers every map surface (FCA map, facility map, replay) in one place, not a partial fix.

**Non-blocking finding** (also independently confirmed): `pointer-events` is CSS-inherited, so any future interactive `<Marker>` passed via `mapChildren` would silently inherit `pointer-events: none` from this fix with no visible error. Traced every current `mapChildren` consumer — only `AtcMarkers` (internal), which already sets `pointer-events: none` on its own badges by design (hover is handled via a separate deck.gl pickable layer per its own comment) — so there's no current regression. But the fix's justification comment ("No native MapLibre control is used") doesn't account for this risk surface at all, and the next contributor adding a clickable/draggable marker via `mapChildren` will hit a silent, hard-to-diagnose dead end. Worth a one-line comment addition warning about this before it bites someone — not requesting rework over it.

**Ran:** `pnpm --filter web typecheck` clean, `pnpm --filter web test` 88/88 passed (unchanged — no regression). No backend/API contract touched (single file, `MapCanvas.tsx`).

**Could not live-browser-verify the actual click restoration** — no browser tooling available in this review environment either (matches the PR's own stated limitation). Confidence is high based on the verified DOM/CSS mechanism above, but the manual click-test the PR itself requests before merging still stands — please do it.

**Verdict:** Correctly targeted, minimal fix for an active regression affecting every map. Approved by operator.

Issue #143, opened 2026-09-12
Build out the Discord bot: restructure into a proper crate + close gaps (templates, RAW TMIs, ACE DMs + reminders, multi-guild)


Vainnor:
Epic: restructure the `ois-discord` bot into a proper crate and close the gaps between what's built
today and the target behaviour below. This is **L/XL — split into the sub-issues proposed at the
end** during triage (board supports Parent/Sub-issues).

## Current state (what's actually built)

The docs still call the bot *"designed"* (`docs/README.md:34,44`), but it's **partially built**:
`discord/src/main.rs` is a single **762-line** file. Architecture (correct, keep it): the backend
enqueues `integration.outbound_jobs`; the bot **drains the queue and calls back via `ois-client` as a
service account — no DB access** (`docs/PLAN.md:18`, `docs/features/discord-integration.md`).

Implemented today (`perform_job` dispatch, `main.rs:420`):
- `ace_request_post` → `post_ace_request` — posts an ACE request embed with a **Claim** button.
- `ace_request_notify` → `notify_ace_claim` — edits that message's embed/button as slots fill.
- `tmi_publish` → `post_tmi` — posts a TMI as a **structured embed**.
- `event_thread_create` → `create_event_thread` — makes a planning thread, pings each facility's
  **EC(s)** (OIS users with the ARTCC-scoped `EC` role, by linked Discord id) + NTMO/DCC-Trainee
  config roles, with availability buttons.
- `guild_snapshot` → `snapshot_and_push` — pushes **every** guild's channels/roles to the backend so
  the website config can offer dropdowns.

Backend/website side already exists: `handlers/integration.rs` (`get/put_discord_config`,
`push/refresh_guild_snapshot`, `discord_ace_info`, `discord_ace_claim`, `discord_availability`),
per-guild config models (`DiscordGuildConfig*`), account linking (`DiscordLinkBody`), and
`events::publish_event_discord`. So **guild/channel config is already website-driven** and the ACE
claim already reflects back to the site.

**Docs are stale** — reconcile `docs/features/discord-integration.md` (+ `ace-support.md`,
`tmu-ntml-adv-tmi.md`, `events-workflow.md`, `README.md` "designed" note) with what's built and with
the target below.

## Target behaviour & gaps

1. **Two guilds (DCC + VATUSA), more later.** One bot application already sees every guild it's in
   (`snapshot_and_push` iterates all `guilds`) and config is per-guild. **Mostly there** — verify
[...]

Issue #148, opened 2026-09-12
No config/storage home exists for per-facility documents (needed for ACE-claim DM)


Vainnor:
## What happens

There is no configuration or storage home for per-facility documents (SOPs, LOAs, etc.) that should
be DMed to a controller when they claim an ACE request.

## Why

`discord_ace_claim` and `notify_ace_claim` exist, but nothing models "this facility's set of
documents" anywhere in the backend.

## What should happen

Add a per-facility document set: a new integration config table (facility → ordered list of
{title, url} or similar), an admin/website UI to manage it, and handlers to read/write it. This is a
prerequisite for sub-issue E (DM on claim).

## Acceptance

- [ ] A facility can have zero or more configured documents (title + link).
- [ ] Website UI to add/edit/remove a facility's documents.
- [ ] A handler/repo function returns a facility's documents for the DM job (sub-issue E) to consume.

Blast radius: backend (new migration + repo + handlers) + web (config UI) + API contract (client
regen). No trajectory/permissions impact beyond a new admin-gated config surface.
Part of the #143 epic (sub-issue D). Relates to #24, #59 (both closed; this replaces their config
half).

🤖 Drafted by Claude Code

Vainnor:
Shipped in #155.

**Files:** `backend/migrations/0064_facility_documents.sql` (new table + 2 permissions), `backend/src/{repos,handlers}/facility_documents.rs` (new), `crates/ois-core/src/catalog.rs` + `backend/src/auth/permissions.rs` (three-in-sync), `backend/src/{router,openapi,models/mod}.rs`, `web/src/lib/facility-documents.ts` + `web/src/pages/planning/facility-documents.tsx` (new), nav + router wiring.

**Blast radius:** new migration (additive) + new permission pair (facility-scoped, three-in-sync verified by the automated test) + new endpoint (API contract change, client regenerated in this PR). No trajectory/ETA model, no existing endpoint changed.

**Deploy note:** migration `0064` applies on backend startup (additive, no backfill). Generated client committed as part of this PR.

Self-review caught two real issues before pushing: `delete` wasn't facility-scoped (cross-facility delete via id), and the web page checked the wrong read permission. Both fixed with a regression test for the first.

Unblocks sub-issue E (DM-on-claim), which will read from `repos::facility_documents::list_by_facility` directly.

PR #157, opened 2026-09-12
feat(discord): website-configured event-thread message template

### LINKED ISSUE #146: Discord event-thread message body is hard-coded; make it a website-configured template
## What happens

`create_event_thread` (`discord/src/main.rs:~665`) builds the event-thread message body from a
hard-coded Rust `format!` block. Any wording change requires a code change + redeploy.

## Why

No website-side config exists for the thread message template; only the guild/channel/role targets
are config-driven.

## What should happen

Add a website-configured event-thread message template (placeholders for event name, time, host
ARTCC, etc.), stored per the existing `DiscordGuildConfig*` pattern. `create_event_thread` renders
the configured template instead of the hard-coded block. Keep the data-driven host + neighbour EC/
role pings unchanged. Confirm the `facilities` payload from `publish_event_discord` includes
neighbour ARTCCs, not just the host, while implementing.

This supersedes the now-closed #50 (deferred pending the bot's actual config/API surface — which now
exists per the epic's "current state" section).

## Acceptance

- [ ] Thread message body comes from a website-configured template, not a Rust `format!` block.
- [ ] Host + neighbour ARTCC EC/role pings still fire correctly.
- [ ] A template change takes effect without a bot redeploy.

Blast radius: backend (`integration`/`events` config + a new migration for the template field) + web
(config UI) + `ois-discord` (render the template instead of hard-coding it) + API contract (client
regen for the new config field).
Part of the #143 epic (sub-issue B). Relates to #50 (superseded).

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Replaces the hard-coded Rust `format!` block in `discord/src/jobs/thread.rs`'s `create_event_thread` with a `{{placeholder}}` template stored in a new singleton table (`integration.event_thread_template`), editable from the existing Discord admin page (`/admin/discord`). Reuses `discord.config.read`/`update` — no new permission.
- The backend fetches the configured template and adds it to the `event_thread_create` job payload; the bot substitutes the same 5 values it already computed (`title`, `date_line`, `facility_lines`, `ntmo_ping`, `dcc_ping`). The seeded default renders byte-identical to the previous hard-coded output.
- Confirmed `publish_event_discord`'s `facilities` payload was already built from required/preferred facility-support rows (host or neighbour, whichever are flagged) rather than a host-only list — no code change needed there for AC2.

## Test plan
- New backend test: `render_template` output for the fallback template, with sample values, asserted equal to a manual reconstruction of the old hard-coded `format!` output — passes, confirming no behavior change on day one.
- New `#[sqlx::test]`s: the seeded default is fetched correctly; set-then-get round-trips.
- `just ci` full gate green (170/170 backend tests, 2 new bot-side tests, clippy, fmt, web typecheck, 88/88 web tests).
- Client regenerated via the `dump_openapi_json` CI utility + `openapi-typescript` (no live server needed).
- Can't click through the admin UI or actually post to Discord in this environment (no display, no bot connection) — a manual save-and-post test is worth doing before merge.

Closes #146 (sub-issue B of epic #143)


Vainnor:

Tested in a worktree on `feat/146/thread-template`. `/code-review` + independent verification surfaced real correctness bugs; fixes are pushed as commit `8b85707` on top of the original `8205553`.

**Findings and fixes:**

1. **Placeholder re-injection** (`discord/src/jobs/thread.rs`) — `render_template` did sequential per-key `.replace()` calls over the *accumulating output*. A value containing literal `{{other_key}}` text (e.g. an event titled `Fly-in {{ntmo_ping}} Weekend` — `event.title` is settable by anyone with `EventsPlanUpdate`, a broader group than the `DiscordConfigUpdate` holders who edit the template) got re-matched and substituted by a later iteration, injecting a duplicate role ping. Verified this could only duplicate a ping *already* going out in that message (both `ntmo`/`dcc` role ids are already in `allowed_mentions` regardless) — not a privilege escalation, but a real bug crossing a privilege boundary. Rewrote as a single left-to-right pass that never feeds substituted values back through the scan, making the whole class structurally impossible. Confirmed via mutation test: reverted to the old sequential-replace implementation, the new regression test failed exactly as expected (`<@&222> Weekend Ping: <@&222>` — duplicate ping), restored the fix, test passes.

2. **No length cap** (`backend/src/handlers/integration.rs`) — only emptiness was checked, nothing against Discord's ~2000-char message limit, and the per-event `facility_lines` block (one line per facility) can push a short template over the edge. Added a 1500-char save-time cap plus a `truncate()` safety net on the final *rendered* content in the bot itself (matching the existing `truncate(thread_name, 100)` pattern already used in this file).

3. **Hard 500 on a missing template row** (`backend/src/repos/integration.rs`) — `fetch_one` would fail `publish_event_discord` entirely (blocking *all* event-thread creation) if the singleton row were ever missing, unlike sibling configs that degrade gracefully. Now falls back to a compiled-in default (same text the migration seeds) with a warning log via `fetch_optional`.

**Verified via mutation test** for finding 1 (shown above). Findings 2–3 covered by new direct unit/`#[sqlx::test]` tests (boundary values, real missing-row-against-real-Postgres deletion) — didn't mutation-test these given their triviality (simple length comparison; `unwrap_or_else` fallback).

**Also noted, not fixed (non-blocking, operator's call):**
- This branch was forked before `#148`'s migration `0064` merged into `next` (migrations jump 0063→0065 in this branch's tree). Not a merge-conflict risk (distinct filenames, and sqlx applies by numeric filename order regardless of git history) — just needs a rebase before merge, standard PR staleness.
- A new singleton table + endpoint pair for one config string, parallel to the existing `discord_configs` aggregate — a legitimate style question, not a defect; the field's unbounded text length arguably justifies a dedicated table over embedding it in the aggregate.
- `set_event_thread_template`'s UPSERT-then-SELECT isn't transactionally linked, so two concurrent saves could show one admin's toast the other's content — extremely low real-world likelihood (same singleton row, near-simultaneous edits), not fixed.
- No validation that a saved template contains all expected placeholders (a typo'd or omitted `{{ntmo_ping}}` silently ships) — self-inflicted admin footgun, not fixed.

**Re-ran full `just ci` after fixes:** 174/174 backend tests (4 new), 5/5 discord tests (3 new), clippy clean, fmt clean, web typecheck clean, zero client-contract drift (regenerated and diffed against committed schema).

**Verdict:** Rework applied and verified; branch pushed. Please do the manual save-and-post test the PR itself already requests, and consider rebasing onto current `next` before merge (picks up `#148`'s `0064` migration).

Vainnor:
## QA review — SECOND PASS, found a regression in my own first fix, fixed and pushed

Per the operator's request, did a genuinely fresh `/code-review` pass on the already-once-reworked branch rather than assuming my first fix was complete. It found real issues — including a regression my own first fix introduced. Fixed as commit `c407df8` on top of `8b85707`.

**Findings and fixes:**

1. **My own truncation safety net could silently eat the ping section** (`discord/src/jobs/thread.rs`) — the 2000-char safety-net truncate I added in the first fix applies to the *fully rendered* content, and `{{ntmo_ping}}`/`{{dcc_ping}}` sit *after* the variable-length `{{facility_lines}}` block in the template. An event with enough required/preferred facilities could push the render past 2000 chars, and blind tail-truncation would cut the NTMO/DCC ping off entirely — `send_message` still succeeds, so nothing errors or logs, but nobody actually gets pinged. This is the exact class of silent-info-loss bug the original PR was written to avoid, reintroduced by my own defensive fix. Fixed by capping `facility_lines` itself (the one genuinely unbounded piece) *before* substitution, so everything after it in the template is never at risk. **Mutation-tested**: reverted the new `cap_facility_lines` to a no-op, the new end-to-end test reproduced the exact bug (visible in the failure output — the ping mention literally cut off mid-ID) and failed as expected; restored the fix.

2. **Byte-length cap vs. character count mismatch** (`backend/src/handlers/integration.rs`) — `MAX_TEMPLATE_LEN` checked `payload.body.len()` (UTF-8 bytes), but the shipped default template itself uses multi-byte box-drawing dividers and emoji. A template that looks well under 1500 characters in the admin's textarea could exceed 1500 *bytes* and get a bare, unexplained 400. Switched to `.chars().count()`. New test proves it with the same kind of multi-byte content as the real default template.

3. **Redundant INSERT-then-SELECT** (`backend/src/repos/integration.rs`) — `set_event_thread_template` did an upsert then a separate read-back just to return the value it already wrote. Now a single query using `RETURNING body`.

**Also raised, not fixed (pre-existing or genuinely low-priority, operator's call):**
- No validation that a saved template contains required placeholders (same gap I flagged in the first pass — an admin could still delete `{{ntmo_ping}}` from the template and it'd save fine).
- `ThreadTemplateCard`'s `useEffect` re-seeds local edit state on every background refetch with no dirty-check — a tab-away-and-back could discard an admin's unsaved edit. Pre-existing in the original PR, not something my rework touched.
- The default template text is now hand-duplicated in three places (migration seed, backend fallback, bot fallback) instead of two — an accepted trade-off for fixing the hard-500 issue across two crates that don't share dependencies.
- `truncate()`'s char-count (not UTF-16-unit) measure is technically imprecise against Discord's actual limit in extreme edge cases — pre-existing in the shared utility, used elsewhere in the codebase the same way.
- Three sequential un-parallelized DB reads in `publish_event_discord` — minor latency, not correctness.

**Re-ran full `just ci`:** 175/175 backend tests (2 new), 8/8 discord tests (3 new), clippy clean, fmt clean, web typecheck clean, 88/88 web tests. No API contract change.

**Verdict:** The first fix was necessary but incomplete — this second pass closes the gap it left open. Recommend a third look isn't needed, but the manual save-and-post test remains worth doing before merge.

PR #161, opened 2026-09-12
feat(discord): post TMIs raw with an ephemeral structured view

### LINKED ISSUE #147: Discord TMI posts always render as structured embeds; add raw post + ephemeral structured view
## What happens

`post_tmi` (dispatched from `tmi_publish`) always posts a TMI to the NTML channel as a structured
Discord embed, even when the controller entered it as free-form raw text.

## Why

The job only builds and posts a structured embed; there's no raw-text posting path and no way to
view the structured breakdown separately.

## What should happen

Post the TMI's **raw** text form to the NTML channel (already config-driven), with a **"View
structured"** button that replies **ephemerally** (visible only to the clicker) with the structured
breakdown. This keeps the channel readable in the controllers' own words while still offering the
structured view on demand.

## Acceptance

- [ ] TMI posts land in the NTML channel as raw text, not a structured embed by default.
- [ ] A "View structured" button posts an ephemeral reply with the structured breakdown.
- [ ] NTML channel target still resolves from website config (unchanged).

Blast radius: `ois-discord` (`jobs/tmi.rs` post + new interaction handler). No backend contract
change if the structured data is already in the job payload; verify during implementation.
Part of the #143 epic (sub-issue C).

🤖 Drafted by Claude Code

Vainnor:
## Summary
- `post_tmi` now posts the TMI as plain raw text (prefixed with "REQ → PROV" for context) instead of a structured embed, with a "View structured" button.
- `TmiBody` already carried everything needed — `restriction` is always the raw NTML line, `structured`/`decoded` are the plain-English breakdown (null for a raw-typed TMI) — no schema change.
- The button calls back into the API at click time via a new bot-callback endpoint (`discord_tmi_info`, mirroring the existing `discord_ace_info` pattern exactly — same `IntegrationJobsUpdate` gate, same `OisClient` shape) since Discord's 100-char `custom_id` limit can't carry the full decoded text.

## A regression I caught and fixed before pushing
My first pass posted only the raw `restriction` line — but `restriction`/`tmi::encode()` never includes the requesting/providing facilities (confirmed by `backend/src/tmi.rs`'s own doc comment: "The requesting/providing facilities... live on the TMI row, not in the line here"). That would have silently dropped information the old embed showed explicitly. Fixed by prefixing the post with the facilities when both are present.

## Test plan
- New bot-side unit tests for `structured_reply`: returns `decoded` when present, a clear fallback message for a raw-typed TMI (no structured data) when absent.
- New backend `#[sqlx::test]` (`repos::tmu::tests::get_tmi_reflects_raw_vs_structured_entry`): creates one raw-typed and one structured-typed TMI, confirms `structured`/`decoded` are null vs. populated as expected.
- `just ci` full gate green (170/170 backend tests, 2 new bot-side tests, clippy, fmt, web typecheck).
- Client regenerated via the `dump_openapi_json` CI utility (bot-only endpoint, not consumed by the web app, but keeps `client-drift` CI green).
- Can't post to a real Discord channel or click the button in this environment — the callback logic is covered by the backend + bot tests; a manual publish-and-click test is worth doing before merge.

Closes #147 (sub-issue C of epic #143)


Vainnor:

Tested in a worktree on `feat/147/raw-tmi-post`. `/code-review` + independent verification found a real security-relevant regression and a correctness gap; fixes pushed as commit `640d8ca` on top of `8f5fbe8`.

**Findings and fixes:**

1. **Missing `allowed_mentions` restriction — mass-ping vector** (`discord/src/jobs/tmi.rs`). The old structured embed is never parsed for mentions by Discord (embeds are immune to this by design). The new plain-content post had no `allowed_mentions` set — verified in serenity's actual `CreateMessage` source that when omitted, it falls back to `http.default_allowed_mentions`, which this bot never configures anywhere, so Discord's own default applies: parse and ping *everything* in the content. Any `TmuTmiCreate` holder (a meaningfully broad TMU-staff population, not just admins) entering a raw restriction containing a literal `@everyone`/`@here` would have caused a real mass-ping of the TMU channel — something structurally impossible with the embed it replaces. Fixed with `.allowed_mentions(CreateAllowedMentions::new())` (denies everything), matching the pattern `jobs/thread.rs` already uses for its own free-text-adjacent content. Not unit-testable (a builder call, not a pure function) — verified correct by reading serenity's source rather than by test.

2. **Facility header used AND logic, dropping both fields when one was missing** (`discord/src/jobs/tmi.rs`) — the embed it replaces showed `requesting`/`providing` independently. `update_tmi` (pre-existing, unrelated to this PR) allows blanking either to empty without rejection, unlike `create_tmi`'s validation — so a TMI edited mid-flow could have a still-known facility silently dropped from the post entirely. Extracted into `facility_header()` and fixed to show whichever field is present. **Mutation-tested**: reverted to AND-only logic, the new test failed exactly as expected (`expected Some("**ZDC**"), got None`); restored the fix, test passes.

3. **Generic error message for every failure, including a 404 that can never succeed on retry** (`discord/src/interactions/tmi_structured.rs`) — `prune_history` hard-deletes published TMI rows past retention, but a Discord message's button can outlive the row. A click on an old post now gets "This TMI is no longer available" instead of the same unhelpful "try again" as a transient 500, mirroring `ace_claim.rs::claim_error`'s status-code dispatch pattern exactly.

4. **`create_response`'s Result silently discarded** — now logged on failure (`tracing::error!`), matching the codebase's primary convention for this call (used everywhere except one other ephemeral-error-reply outlier).

**Not fixed (non-blocking, operator's call):**
- No length cap on the ephemeral reply vs. Discord's 2000-char limit for a very long free-text (`kind: "TXT"`) structured restriction — plausible but narrow edge case.
- No status filtering on `discord_tmi_info` for expired/cancelled TMIs — showing historical data for an old click isn't clearly wrong.
- `DiscordTmiInfoBody`/`DiscordTmiInfo` duplicate `TmiBody`'s shape by hand — consistent with the existing `DiscordAceInfoBody` precedent, not a new pattern this PR introduced.
- Minor DRY nits (duplicated `seed_user` test helper, no shared "add details button" convention across job types) — cosmetic, not fixed.

**Re-ran full `just ci` after fixes:** 169/169 backend tests (unchanged), 6/6 discord tests (4 new), clippy clean, fmt clean. No API contract touched by this rework (client drift unaffected).

**Verdict:** Rework applied and verified; branch pushed. Please do the manual publish-and-click test the PR itself already requests.

Vainnor:
## QA review — SECOND PASS, no new issues in the rework itself

Per the operator's request, did a genuinely fresh `/code-review` pass on the already-reworked branch. Unlike #146's second pass, this one found **zero issues in what I actually changed** — the mass-ping fix (`CreateAllowedMentions::new()`) and the facility-header degradation fix both hold up cleanly.

**Re-verified independently:** fresh `just ci` (169/169 backend, 6/6 discord), fmt, clippy — all clean, matching the first pass exactly.

**Two findings surfaced, both pre-existing/out-of-scope, not new regressions:**

1. **No length cap on `restriction`** — `backend/src/handlers/tmu.rs::create_tmi` only trims and empty-checks `restriction`, no max length. If a TMU controller entered an unusually long raw restriction, `post_tmi`'s `send_message` would fail outright against Discord's 2000-char content limit — a logged job failure, not a silent one. This gap predates #147 entirely (it's in `create_tmi`, which #147 doesn't touch) and the *old embed* this PR replaces had the same exposure (embed field limits would have rejected an oversized restriction too). Worth a separate follow-up issue for `create_tmi`/`update_tmi` input validation, but not something #147 introduced or should be blocked on.

2. **"View structured" interaction isn't deferred before its HTTP round-trip** — `tmi_structured.rs::handle` awaits `api.tmi_info(tmi_id)` before calling `create_response`, risking Discord's ~3-second interaction-ack window on a slow backend. The reviewer explicitly confirmed this "mirrors an existing await-then-respond pattern elsewhere in the bot" (e.g. `ace_claim.rs`'s `handle_claim_button`) — a pre-existing bot-wide convention, not a new anti-pattern this PR introduces.

**Verdict:** The mass-ping and facility-header fixes are solid under independent re-scrutiny. The two informational findings are legitimate but belong in separate follow-up issues (input validation on TMI restriction length; interaction-deferral convention across the bot), not this one. Approved by operator.

PR #163, opened 2026-09-12
fix(map): align deck.gl layers/extensions to 9.4, matching core/react

### LINKED ISSUE #159: Aircraft unclickable + map hover tooltips gone — deck.gl version skew (core/react 9.4 vs layers/extensions 9.3.10)
## What happens

On the FCA map (and every map): **clicking an aircraft does nothing**, and **hover tooltips no longer
appear**. FCA **lines are still clickable**, so deck.gl is receiving pointer events — it's the
aircraft/label/hover layers that stopped picking. Regression from the recent map dependency bumps.

## Why (root cause — deck.gl version skew)

deck.gl requires **every `@deck.gl/*` (and `@luma.gl/*`) package at the same version**. Dependabot
bumped `core` and `react` to 9.4 (#122, #121) but **not** `@deck.gl/layers` / `@deck.gl/extensions`,
so they're mismatched. On `main`:

`web/package.json`:
```
@deck.gl/core       ^9.4
@deck.gl/react      ^9.4
@deck.gl/layers     ^9.1   ← not bumped
@deck.gl/extensions ^9.1   ← not bumped
```

Resolved in `pnpm-lock.yaml`:
```
@deck.gl/core@9.4.0   @deck.gl/react@9.4.0
@deck.gl/layers@9.3.10   @deck.gl/extensions@9.3.10   (@luma.gl/*@9.4.1)
```

So `IconLayer` + `TextLayer` (from `@deck.gl/layers@9.3.10`) run against `@deck.gl/core@9.4.0` +
`@luma.gl@9.4.1`. That mismatch breaks the parts of the picking pipeline those layers use:
- **Aircraft** = a `pickable` `IconLayer` (`web/src/components/map/layers/aircraft.ts:34-45`, masked
  icons) → clicks don't resolve, so `TrafficMap.tsx:221` (`info.layer?.id === "aircraft"`) never fires.
- **Hover tooltips** = deck `getTooltip` (`TrafficMap.tsx:269` `mapTooltip`) picking the hovered
  object, incl. the ATC hover `IconLayer` (`buildAtcHoverLayer`) → hover picks return nothing.
- **FCA lines** are a path/line layer and still pick, which is why they work — consistent with the
  skew hitting the icon/text layers, not path layers.

(The earlier basemap-passthrough fix #156 restored *line* clicks by letting events reach deck; it did
not — and could not — fix this, which is a separate dependency-version defect.)

## What should happen

Align the whole deck.gl stack to one version: bump `@deck.gl/layers` and `@deck.gl/extensions` to
`^9.4` (matching `core`/`react`), so they resolve to `9.4.0`, and refresh `pnpm-lock.yaml`. Then all
`@deck.gl/*` are 9.4.x and `@luma.gl/*` are 9.4.x, and IconLayer/hover picking works again.

## Acceptance

- [ ] All `@deck.gl/*` packages resolve to the same 9.4.x version in `pnpm-lock.yaml` (no 9.3.x
      layers/extensions against 9.4 core).
- [ ] Clicking an aircraft on any map selects it / focuses its callsign.
- [ ] Hover tooltips (aircraft + ATC) show again; FCA line clicks still work.

Blast radius: web only (`web/package.json` deck.gl versions + lockfile). No backend, no API contract.
Introduced by the partial deck.gl 9.4 bump (#121/#122 raised core/react; layers/extensions left at
9.1). Related to the earlier map-click work #154/#156 but a distinct root cause. Likely on `main` now.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- `@deck.gl/layers` and `@deck.gl/extensions` were still pinned to `^9.1` while `core`/`react` were bumped to `^9.4` (#121/#122), resolving to a mismatched `9.3.10`/`9.4.0` split — deck.gl explicitly requires every `@deck.gl/*` package at the same version. Confirmed the mismatch directly in `pnpm-lock.yaml` before making any change.
- That skew broke the `IconLayer`/`TextLayer` picking pipeline (aircraft clicks, hover tooltips) while path/line layers (FCA routes) were unaffected — consistent with the reported symptom.
- All four packages now resolve to `9.4.0`, `@luma.gl/*` to `9.4.1`. No source changes needed — checked deck.gl's real changelog for 9.1→9.4, no documented breaking changes to the layers this app uses (`IconLayer`, `TextLayer`, `PathLayer`, `PolygonLayer`, `GeoJsonLayer`, `ScatterplotLayer`).
- A pre-existing, unrelated peer-dependency notice for `@deck.gl/widgets` (9.3.10 vs core's 9.4.0) is untouched — confirmed it already exists on the current unmodified lockfile too (not introduced by this PR), and the package isn't imported anywhere in this codebase (deck.gl's optional UI controls, unused here).

## Test plan
- Confirmed in the regenerated lockfile that `@deck.gl/core`/`react`/`layers`/`extensions` all resolve to `9.4.0` (the actual AC1).
- Can't click an aircraft or hover a tooltip in this environment (no browser) — the version alignment is the fix per the issue's own diagnosis; a manual click/hover test is worth doing before merge (same caveat as #154/#156).

Closes #159


Vainnor:
## QA review — PASS

Tested in a worktree on `fix/159/deckgl-version-skew`.

**Verified independently:**
- Confirmed `pnpm install --frozen-lockfile` succeeds (no drift between `package.json` and the committed `pnpm-lock.yaml`).
- Grepped the resolved lockfile directly: `@deck.gl/core`, `@deck.gl/react`, `@deck.gl/layers`, `@deck.gl/extensions` all resolve to exactly `9.4.0` — the actual AC1.
- Ran a full production build (`pnpm build`) in addition to typecheck — succeeded cleanly (the one chunk-size warning is pre-existing/unrelated to this bump).
- `pnpm --filter web typecheck` clean, `pnpm --filter web test` 88/88 passed.

**`/code-review high`** flagged that `@deck.gl/widgets` remains at `9.3.10` — the same skew class this PR fixes for `layers`/`extensions`, since `@deck.gl/react@9.4.0` declares a non-optional peer on `~9.4.0`. I verified: this is **pre-existing** (present in `origin/next`'s lockfile before this PR, not introduced by it), already **explicitly disclosed** in this PR's own description, and the package is **not imported anywhere** in this app (confirmed via grep — no `widgets` prop passed to `<DeckGL>`). Currently inert; would only matter if a future feature reaches for a deck.gl widget. Worth a small follow-up issue, but out of scope for what #159 asked for (core/react/layers/extensions alignment only) and not a regression this PR introduces.

**Verdict:** Correctly scoped, verified fix for the reported regression. Approved by operator.

Issue #2, opened 2026-08-19
fine grained aircraft profiles


Vainnor:

Vainnor:
**Implemented** ✅ — shipped in commit `75caafb` ("feat(flow): configurable aircraft performance profiles").

Covers: a vertical-profile ETA model (climb/cruise/descent schedules, ISA Mach↔TAS, real top-of-descent, service-ceiling cap), a `flow.aircraft_profile` catalog resolved by exact type → wake class → default, applied across FCA metering / airport-flow demand / runway ETE, a CRUD API gated by `flow.aircraft_profiles.*`, and a Planning → Aircraft profiles editor page.

Safe to close.

Issue #1, opened 2026-08-17
Dashboard Arrival Ladder Configuration


Vainnor:
Add the ability to add different filters to the ladder configuration in the dashboard

Issue #149, opened 2026-09-12
ACE claim doesn't DM the claimer with facility documents or a confirmation


Vainnor:
## What happens

When a controller claims an ACE request, `notify_ace_claim` only edits the original channel
message/button — the claimer gets no DM, no facility documents, and no "you're working this event"
confirmation.

## Why

No DM job type exists in `ois-discord` today; `perform_job` has no arm for it, and there's no
outbound-job enqueue for it in the backend's claim handler.

## What should happen

Add a new outbound job type: on `discord_ace_claim`, enqueue a DM-the-claimer job carrying the
claimed event's facility-configured documents (sub-issue D) plus a confirmation message. `ois-discord`
gets a new `jobs/dm.rs` handling this job type via the existing queue-drain + `ois-client` callback
design (no DB access from the bot).

Depends on sub-issue D (facility documents must exist to send).

## Acceptance

- [ ] Claiming an ACE request enqueues a DM job to the claimer.
- [ ] The DM includes the facility's configured documents (or a note if none configured) and a
      confirmation that they're working the event.
- [ ] A test covers the enqueue-on-claim path (unit or integration, per existing job-enqueue test
      patterns).

Blast radius: backend (`handlers/integration.rs`'s claim path enqueues a new job type; `jobs.rs`) +
`ois-discord` (new `jobs/dm.rs` arm). No API contract change if using the existing outbound-job
envelope; verify during implementation.
Part of the #143 epic (sub-issue E). Relates to #24, #59 (both closed; this replaces the DM half).

🤖 Drafted by Claude Code

Vainnor:
Shipped in #162.

**Files:** `backend/src/handlers/ace.rs` (new `enqueue_claim_dm`, called from both claim entry points), `backend/src/handlers/integration.rs` (wired into `discord_ace_claim`), `discord/src/jobs/dm.rs` (new — DM sender + content builder), `discord/src/jobs/mod.rs` (dispatch wiring).

**Blast radius:** backend (new job type, reuses the existing outbound-job queue — no new permission, no new endpoint) + `ois-discord`. No web change, no migration, no API contract change.

**Deploy note:** none.

**Verification:** new backend `#[sqlx::test]`s covering both the Discord-linked and unlinked-claimer paths, new bot-side tests for the DM content builder. `just ci` green throughout (177/177 backend tests). No Discord connection available here to send/verify a real DM — worth a manual claim-and-check test before merge.

[...]

PR #175, opened 2026-09-12
feat(discord): ACE-claim reminder DMs at T-24h/T-6h

### LINKED ISSUE #150: No reminder DMs are scheduled ahead of a confirmed ACE claim's event (T-24h/T-6h)
## What happens

Once a controller claims an ACE request (sub-issue E), they get a one-time confirmation DM but no
reminder as the event approaches.

## Why

No reminder scheduling exists in `jobs.rs` — outbound jobs are only enqueued at claim-time, not on a
timer relative to the event start.

## What should happen

Backend (`jobs.rs`) enqueues DM reminder outbound jobs at **T-24h** and **T-6h** before the event
start for each confirmed claim, reusing the DM job type/plumbing from sub-issue E.

Depends on sub-issue E (the DM job type must exist first).

## Acceptance

- [ ] A confirmed ACE claim schedules two reminder DM jobs, at T-24h and T-6h before event start.
- [ ] A cancelled/unclaimed event does not send stale reminders (verify cancellation path clears
      scheduled jobs).
- [ ] A test covers the scheduling math (T-24h/T-6h relative to event start) and the cancellation
      case.

Blast radius: backend only (`jobs.rs` scheduling + a migration if reminders need persisted schedule
state). No API contract change expected; verify during implementation.
Part of the #143 epic (sub-issue F). Relates to #58 (closed; this replaces it — note the epic's
target is T-24h/T-6h, not #58's original 24h/1h).

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Adds a periodic scheduler (`backend/src/jobs.rs`, mirroring the existing `spawn_capture_scheduler`/`spawn_event_fca_lifecycle` pattern) that DMs ACE claimers a reminder once their event crosses the T-24h or T-6h threshold, reusing #149's DM job/plumbing (`discord/src/jobs/dm.rs`).
- **Idempotent by construction**: each 15-minute tick re-queries live state (crossed the threshold, event still upcoming, request not `cancelled`, not already reminded via a `not exists` check against `integration.outbound_jobs`) rather than pre-scheduling into the future. A released claim (`release_claim` deletes its `ace.claims` row) or a cancelled request simply stops matching on the next tick — no separate "clear the scheduled reminder" step needed, and no new migration.
- `build_dm_content` gains a `reminder` branch producing a short line without re-listing documents (already sent in the original confirmation).

## Test plan
- New backend `#[sqlx::test]`s on `claims_due_for_reminder`: matches the T-24h scheduling math (23h-out event due, 30h-out not); excludes a cancelled request even though its event is in-window; is idempotent (an already-enqueued reminder doesn't resurface).
- New bot-side unit tests for the reminder content branch (with/without a position).
- `just ci` full gate green (188/188 backend tests, 19/19 bot-side tests, clippy, fmt, web typecheck).
- No migration, no new endpoint, no API contract change.
- Can't observe a real scheduled DM fire in this environment — the query/idempotency logic is covered by the backend test, the message body by the bot test; a manual end-to-end check (claim something with an event ~24h out, wait for a scheduler tick) is worth doing before merge.

Closes #150 (sub-issue F of epic #143)


Vainnor:

Tested in a worktree on `feat/150/ace-reminders`. Independently verified the core design (dual job_type dedup, cancellation/release exclusion) before `/code-review` ran — both held up. `/code-review` then surfaced 9 candidates; 3 were real and worth fixing, pushed as commit `523ae5a` on top of `78225e9`.

**Findings and fixes:**

1. **No lower time bound → wrong "24h" label + double DM for late claims** (`backend/src/repos/ace.rs`) — `claims_due_for_reminder` only bounded time-to-event from above (`<= now() + N hours`), never below. A claim first seen after its event was already inside a *later* tier's window (e.g. a same-day claim made 3 hours before the event, before either reminder fires) matched **both** the 24h and 6h tier queries on the same tick, since 3h ≤ 24h and 3h ≤ 6h are both true and neither job_type existed yet. The claimer would get two DMs back-to-back, one factually wrong ("in about 24h" for an event 3 hours out). Fixed: each tier now queries its own bounded window — `(6h, 24h]` and `(0h, 6h]` — so a claim only ever matches the tier it's genuinely, currently in. **Mutation-tested**: reverted the lower bound, the new regression test reproduced the exact bug and failed; restored.

2. **TOCTOU race with no DB-level backing** (`backend/src/repos/ace.rs`, new migration `0066`) — the SELECT-then-INSERT dedup pattern has no unique constraint, so two backend processes briefly overlapping during a routine rolling deploy (not just permanent multi-replica) could both see a claim as "not yet reminded" and both enqueue a duplicate reminder. Added a **partial** unique index scoped to just the two reminder job types — deliberately *not* a blanket constraint on `(job_type, subject_type, subject_id)`, since other job types sharing `outbound_jobs` (e.g. `ace_request_notify`) legitimately enqueue more than once per subject over time (once per claim/release), which a general constraint would silently break. New `enqueue_reminder_job()` does the insert with `on conflict ... do nothing`, atomically closing the race the old select-then-insert couldn't. **Mutation-tested**: removed the `ON CONFLICT` clause, the atomic-insert test failed with a real constraint violation (proving the index is genuinely enforced, not just present); restored.

3. **One tier's transient failure aborted the whole scheduler pass** (`backend/src/jobs.rs`) — the `?` on the first tier's query propagated immediately, skipping the independent second tier's check for that entire 15-minute cycle. Each tier's query/enqueue is now isolated — a failure is logged and the loop continues to the next tier. Also fixed a silently-swallowed `pool.begin()` failure with no log line (now matches the enqueue-failure arm right beside it).

**Also raised, not fixed (non-blocking, operator's call):**
- Dedup check matches on mere row existence regardless of status, so a permanently-`failed` DM (Discord DMs disabled/blocked) silently suppresses that claim's reminder for that tier forever, with no retry path. Inherent to any DM-based delivery on Discord's opt-in privacy model; the existing job-retry system already retries transient failures up to `MAX_ATTEMPTS`. Building guaranteed-delivery/admin-retry infrastructure would be over-engineering for a best-effort reminder feature.
- `job_type` strings duplicated between `backend/src/jobs.rs` and `discord/src/jobs/mod.rs`'s dispatch match arm (two crates, no shared dependency) — same accepted cross-crate trade-off as #146's fallback-template duplication.
- Real sqlx errors discarded via `.map_err(|_| ApiError::Internal)` — this is the codebase's universal, pre-existing convention in literally every repo function reviewed so far, not something unique to this PR.
- Per-row transaction for a single atomic insert (vs. sibling schedulers calling repos directly against `&PgPool`) — `enqueue_job`/`enqueue_reminder_job` require a transaction by existing signature; adding a non-transactional variant just to shave a negligible per-tick overhead would be over-engineering.
- Two sequential queries per tick instead of one combined query — negligible cost at a 15-minute cadence.

**Re-ran full `just ci` after fixes:** 187/187 backend tests (2 new), 19/19 discord tests (unchanged), clippy clean, fmt clean, web typecheck + 88/88 web tests unaffected. New migration is additive (`create unique index if not exists`) and applies cleanly per `migrations_apply_cleanly`.

**Verdict:** Rework applied and verified; branch pushed. Please do the manual claim-and-wait end-to-end test the PR itself already requests.

Vainnor:
**Second-pass review findings — fixed:**

1. **Masked persistent tier failure** — `ace_reminder_scheduler_once` returned `Ok` for the whole cycle as long as `sent > 0`, even when one tier's query was *persistently* failing (only masked if the *other* tier had hits that cycle). Changed `if tier_failed && sent == 0` → `if tier_failed` so any tier failure always surfaces to the JobRegistry, independent of the other tier's outcome.
2. **Needless transaction wrapper** — `enqueue_reminder_job`'s single `INSERT ... ON CONFLICT ... RETURNING` is already atomic; the caller no longer needs to `pool.begin()`/`tx.commit()` around it. Simplified signature to `&PgPool`, updated the one existing test to match.

Verified: `cargo test -p ois-backend` (187 passed), `just ci` (backend + discord + web, all green), `cargo fmt --check`, `cargo clippy --all-targets -D warnings` — all clean. Fix #1 is a one-line boolean-condition change verified by inspection (triggering it requires a genuine DB query failure, which isn't practical to simulate in a `sqlx::test` integration test); fix #2 is covered by the existing `enqueue_reminder_job_is_atomic_against_a_duplicate_insert` test, updated to the new signature.

Pushed to `feat/150/ace-reminders`.

PR #187, opened 2026-09-12
feat(web): live updates + surfaced errors on the Background Tasks page

### LINKED ISSUE #184: Background Tasks page: live updates while a task runs + surface errors (currently a slow 5s poll, trigger errors swallowed)
## What we want

On the **Background Tasks** admin page (`web/src/pages/admin/jobs.tsx`, route `/admin/jobs`), running a
task feels unresponsive. Hitting **Run now** should give prompt, live feedback: the task flips to
*Running…* immediately, updates to *OK*/*Failed* as soon as it finishes (not seconds later), and any
error is surfaced clearly on the page rather than buried or dropped.

## Why it feels sluggish today

The table is driven by a **fixed 5-second poll**, and the trigger does a single delayed refresh:

```ts
// web/src/lib/jobs.ts:17
refetchInterval: 5000,
```
```ts
// web/src/lib/jobs.ts:32 — after triggering a run
onSuccess: () => setTimeout(() => qc.invalidateQueries({ queryKey: ["admin-jobs"] }), 400),
```

So after clicking Run, it can take up to ~5 s for *Running…* to appear and up to another ~5 s for the
final *OK/Failed* — the page looks like nothing happened. The Run button disables on
`job.running || run.isPending` (`jobs.tsx:110`), but `run.isPending` ends when the POST returns, before
the poll catches up, so there's a visible dead gap.

## Why errors don't show up well

- **The trigger error is swallowed.** `useRunJob` throws on a failed POST but nothing consumes it —
  no toast, no inline message (`web/src/lib/jobs.ts:29`). If triggering a job fails, the user gets
  zero feedback.
- **A failed job is under-surfaced.** A failure shows only as a small `destructive` badge
  (`StatusBadge`, `jobs.tsx:75`) plus the `last_detail` text in a muted Detail cell (`jobs.tsx:100`) —
  easy to miss, and the message is truncated to one cell.

## What should happen

A frontend-only enhancement (no job-system change required):

1. **Responsive run feedback.** On trigger, reflect *Running…* immediately (optimistic update of that
   row) instead of waiting for the poll, and keep the button disabled until the run resolves.
2. **Live updates while active.** Make the poll adaptive: refetch fast (~1 s) while any job is
   `running`, and fall back to the current idle cadence (5 s) when nothing is running — so completion
   (and OK/Failed) appears within ~a second, without hammering the endpoint when idle.
3. **Surface errors on the page.** Render the trigger mutation error inline / as a toast
   (`useRunJob`'s error is currently unused), and make a failed job's `last_detail` prominent (e.g.
   error-styled detail, not just a muted cell) so a failure is obvious where the task ran.

Out of scope (note for triage): a true push/streaming approach (SSE/WebSocket for job state) would be
even more live but needs a new backend endpoint — adaptive polling gets ~90% of the feel with a
web-only change. Call it if the team wants the streaming version instead.

## Acceptance

- [ ] Clicking **Run now** shows *Running…* on that task effectively immediately (no multi-second gap).
- [ ] A task's completion (OK or Failed) appears within ~1 s of finishing while the page is open.
- [ ] The poll returns to the idle cadence when no task is running (no constant 1 s polling at rest).
- [ ] A failed **trigger** (the POST itself) shows a visible error on the page, not silence.
- [ ] A failed **task** surfaces its error detail prominently, not only as a small muted cell.

Blast radius: web only (`web/src/lib/jobs.ts`, `web/src/pages/admin/jobs.tsx`). No backend, no API
contract, no trajectory/permissions — unless the team opts into the streaming variant noted above,
which would add a backend endpoint + client regen.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Replaces the fixed 5s poll (`useJobs`) with an adaptive one — 1s while any task is `running`, 5s at rest — using TanStack Query v5's function-form `refetchInterval` (confirmed the type supports this before using it).
- Replaces `useRunJob`'s fragile `setTimeout(..., 400)` guess with a proper optimistic update: the triggered row flips to "Running…" immediately (no poll wait), a failed trigger POST now surfaces via toast instead of being silently thrown away, and `onSettled` invalidates right away so real state lands as soon as the request resolves — the now-fast poll (since a job shows running) picks up the true completion within ~1s.
- A failed task's `last_detail` is now styled as an error (`text-destructive font-medium`) instead of blending into the muted Detail column.

## Test plan
- New `web/src/lib/jobs.test.ts` (5 tests) for the extracted `jobsPollInterval` pure function: fast when running, slow when idle, fast if *any* job is running, and the `undefined`/empty-list edge cases.
- `pnpm --filter web typecheck` clean, `pnpm --filter web test` 93/93 passed (existing `pages/admin/jobs.test.ts` for `formatBytes` untouched and still passing).
- Can't click through the live page in this environment (no browser) — the polling-cadence and optimistic-update logic is unit-tested directly; a manual click-through (trigger a job, watch it flip to Running then resolve within ~1s) is worth doing before merge.

Closes #184


Vainnor:
**Review findings — sent back for rework, now fixed:**

Two independent passes (manual trace + a fresh `/code-review` agent) converged on the same root cause in `useRunJob`:

1. **onSettled's immediate `invalidateQueries` raced the backend.** `JobRegistry::trigger` only calls `notify_one()` and returns 202 before the job loop actually calls `begin()` (which flips `running=true`). The eager refetch right after the POST resolved could land before that, overwriting the optimistic `running: true` with stale `running: false` — reverting the row and dropping the poll back to the slow 5s cadence. That directly undercut this issue's own "no multi-second gap" acceptance criterion.
2. **onError's rollback wasn't scoped.** It restored a full prior-list snapshot captured at that mutation's `onMutate` time, which could clobber a *different* job's own concurrent optimistic update if two jobs were triggered in quick succession and the first one's trigger failed.

**Fix:** dropped the eager invalidate-on-success entirely — the optimistic update already flips the poll to the fast (1s) cadence, so the next natural tick converges on real state without racing. Scoped the error-path rollback to just the mutated job's own row (new `withJobRestored` helper) so concurrent per-row mutations can't stomp each other. Extracted both cache-mutation steps (`withJobRunning`, `withJobRestored`) as pure functions alongside the existing `jobsPollInterval`, with new unit tests — including one that directly reproduces bug #2 (a second job's optimistic state must survive the first job's rollback) and was mutation-tested (reverting the fix reliably fails that test).

**Verified:** `cargo test --workspace` (182 backend + 17 discord + 3 core, all passing), `cargo fmt --check`, `cargo check --workspace`, `pnpm typecheck`, and `pnpm --filter web test` (97/97, up from 93 — 4 new tests) all clean.

Pushed to `feat/184/jobs-page-live-updates`.

Vainnor:
**Third-pass findings — fixed:**

A fresh `/code-review` pass on the previous rework found three more issues in `useRunJob`, all addressed:

1. **Rollback clobbered fresher same-job data.** `onError` restored the entire `previousJob` snapshot, which could discard runs/last_ok/last_detail a concurrent poll had already refreshed for that same job (e.g. the trigger actually succeeded server-side but the client-perceived request failed). Now `onError` patches back only the `running` flag it itself set, via the merged `updateJob(jobs, name, updater)` helper — everything else on the row is left as the cache currently has it.
2. **Same-job double-click race.** Two `mutate()` calls fired before a re-render could disable the button would interleave their `onMutate`/`onError` against the same cached row. Fixed with TanStack Query's built-in `scope: { id: \`run-job:${name}\` } }` — mutations sharing a scope id are queued and run one at a time rather than concurrently. This is a documented library feature, verified by reading the installed v5.101.4 type definitions (`MutationScope = { id: string }`) rather than a runtime test, since it isn't practical to fire two genuinely-overlapping mutate() calls inside a unit test without a hook-rendering harness (none exists in this codebase).
3. **Minor duplication.** Merged `withJobRunning`/`withJobRestored` into one `updateJob` helper.

`useRunJob` now takes the job name as a hook argument (`useRunJob(job.name)`) instead of at `mutate()` time, to give `scope` a stable id.

**Verified:** `cargo test --workspace` (182 backend + 17 discord + 3 core), `cargo fmt --check`, `cargo check --workspace`, `pnpm typecheck` all clean. `pnpm --filter web test`: 97/97 passing, including a new test that directly reproduces finding #1 (mutation-tested: reverting `updateJob` to replace the whole row instead of patching reliably fails it).

Pushed to `feat/184/jobs-page-live-updates`.

Issue #186, opened 2026-09-12
Map rendering pipeline cleanup: coordinate convention, layer memoization, TrafficMap/FcaMapView dedup


Vainnor:
## What happens

#160 combined a concrete bug (stretched ATC/approach boundaries — fixed in #185, narrowed scope)
with a broader "audit the map rendering pipeline" ambition that was deliberately deferred rather
than pursued. This issue preserves that remaining scope so it isn't lost when #160 closes.

## Why

`web/src/components/map/{MapCanvas.tsx, TrafficMap.tsx, layers/*, lib/*}` has grown organically
across many separate features (aircraft, ATC, boundaries, FCA, matched traffic, replay, routes,
draft drawing). #160's own text called out real, still-open concerns:

- **No single documented coordinate convention** across `layers/*`/`lib/geo` — ad-hoc lat/lon swaps
  exist per-file (the #160 fix guarded against *symptoms* of bad geometry, it didn't establish a
  convention that prevents a similar mistake elsewhere).
- **Layer arrays likely aren't memoized** — rebuilding on every render from changing references is
  a common flicker source; not verified either way yet.
- **Duplication between `TrafficMap.tsx` and `FcaMapView.tsx`** where their layer wiring overlaps.
- **Geometry-builder test coverage** exists for `layers/fca.ts` and now `layers/atc.ts` (#185), but
  not the rest of `layers/*`.

## What should happen

Audit and tidy the map rendering pipeline:
- One consistent way to build layers; memoize layer arrays so they aren't rebuilt every render from
  changing references.
- A single, documented coordinate convention (`[lng, lat]`, matching deck.gl's own expectation)
  across `layers/*` and `lib/geo`; kill any ad-hoc lat/lon swaps found along the way.
- Guard remaining polygon/area layers (beyond `atc.ts`, already done) against degenerate input the
  same way #185 did.
- Reduce duplication between `TrafficMap`/`FcaMapView` layer wiring where they overlap.
- Add targeted unit tests for the remaining geometry builders, mirroring `layers/fca.test.ts` and
  `layers/atc.test.ts`.

## Acceptance

- [ ] `layers/*` use a single documented coordinate convention; no stray lat/lon flips.
- [ ] Layer arrays are memoized/stable; no needless per-render rebuilds causing flicker.
- [ ] Remaining geometry builders have unit coverage for the empty/partial/degenerate cases.
- [ ] Duplication between `TrafficMap`/`FcaMapView` is reduced where they overlap.
[...]

Issue #177, opened 2026-09-12
Airport surface-data model + open-source seed (gates/ramps/taxiways)


Vainnor:
## What happens

There is no editable model of an airport's physical surface (gates, ramps/aprons, taxiways). Every
airport-specific ground-movement fact is either hard-coded or absent, which blocks any per-gate/
per-runway estimate work.

## Why

Nothing analogous to `handlers/airport_configs.rs`'s config-domain pattern exists for surface
geometry. Runways already have real data (`feed/runway_db.rs`, OurAirports-sourced ends) — this
extends the same idea to gates/parking positions, ramp/apron areas, and taxiways.

## What should happen

New config domain (migration + models + CRUD handlers), mirroring `handlers/airport_configs.rs`:
- Gates / parking positions — points.
- Ramp & apron areas — polygons.
- Taxiways — lines.
- Runways — reuse the existing OurAirports ends already in `feed/runway_db.rs`; don't duplicate.

Seed from **OpenStreetMap** (`aeroway=gate|parking_position|apron|taxiway`) where available, and
provide an ingestion path for facility **CRC profiles / GeoJSON** where FEs already have them.
Confirm at triage: exact seed-data licensing/attribution requirements for OSM data bundled or
fetched at runtime.

## Acceptance

- [ ] A migration + model set exists for gates/parking positions (points), ramp/apron areas
      (polygons), and taxiways (lines), keyed by airport.
- [ ] CRUD repo + handlers exist (permission-gating is sub-issue B's concern; this issue can gate
      behind a placeholder/existing permission if needed to unblock).
- [ ] At least one real airport's surface data is seeded from OpenStreetMap as a proof of the
      ingestion path.
- [ ] Runways are referenced from the existing `feed/runway_db.rs` data, not re-modeled.

Blast radius: new migration + models + handlers (additive, no existing endpoint changed). No
trajectory/ETA impact yet — that's sub-issue E.
Part of the #164 epic (sub-issue A — do first; everything else keys on this).

🤖 Drafted by Claude Code

Vainnor:
Shipped in PR #189.

**Files:** \`backend/migrations/0067_airport_surface_data.sql\` (new tables + KDCA seed), \`backend/src/{repos,handlers}/airport_surface.rs\` (new), \`backend/src/{router,openapi,models/mod}.rs\` (wiring), \`packages/api-client/src/generated/schema.d.ts\` (regenerated).

[...]

PR #190, opened 2026-09-13
feat(flow): permissioned map editor for airport surface data

### LINKED ISSUE #178: Permissioned map-based editor for airport surface data (gates/ramps/taxiways)
## What happens

Sub-issue A's surface data (gates, ramps, taxiways) needs to be editable by permissioned staff — a
CRUD API alone isn't usable by non-technical facility staff.

## Why

The map already has draft/route-drawing layers (`web/src/components/map/fca/editors.tsx` and
similar) that this can mirror for drawing/labeling points, polygons, and lines.

## What should happen

A map-based editor (mirrors an existing admin page's shell + the map's existing draft-drawing
layers) to draw/label gates, taxiways, ramp areas, and view runways for a chosen airport.
Permission-gated by a **new permission** (e.g. `flow.surface_data.update`) — add it to
`auth/permissions.rs` + `crates/ois-core/src/catalog.rs` + an `access.permissions` migration (the
three-in-sync invariant).

## Acceptance

- [ ] Staff holding the new permission can add/edit/delete gates, ramp/apron areas, and taxiways for
      an airport on a map-based editor.
- [ ] Runways are shown read-only (sourced from `feed/runway_db.rs`, not edited here).
- [ ] The new permission is three-in-sync (marker + catalog entry + migration row) — verified by the
      existing automated sync test (`auth::permissions::sync_tests`).

Blast radius: new permission (three-in-sync) + web (new editor page, mirrors an existing map-drawing
pattern) + the CRUD endpoints from sub-issue A. No trajectory/ETA model impact.
Part of the #164 epic (sub-issue B). Depends on sub-issue A.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Adds the dedicated `flow.surface_data.update` permission (three-in-sync: marker + catalog entry + migration `0068`), swapping sub-issue A's CRUD handlers off the borrowed `events.config.update` placeholder onto it.
- Adds the web editor: `/planning/airport-surface` — a map-based tool to draw/edit gates, ramp/apron areas, and taxiways for a chosen airport, built directly on the generic `MapCanvas` shell (not `TrafficMap`, which drags in unrelated traffic/ATC props).
- Draw/edit state machine is adapted from the FCA polyline editor (`FcaMapView.tsx`), extended to a single-click point mode (gates) and a closable-ring polygon mode (ramp/apron areas), plus the same Enter/Escape/Backspace shortcuts.
- Runways are listed read-only (id/heading/length) via the existing `GET /api/v1/flow/runway/{icao}` endpoint — no new runway code, no geometry (no lat/lon exists anywhere in this codebase for runway ends).

Closes #178

## Test plan
- `cargo test -p ois-backend auth::permissions::sync_tests` and `handlers::airport_surface` — all pass (the sync test is this issue's own AC verification mechanism; the CRUD tests are unaffected by the permission swap since they call the repo layer directly).
- `pnpm --filter web typecheck` and `pnpm --filter web test` — clean, 88 existing tests unaffected.
- Live end-to-end check against an isolated throwaway Postgres (migrations 1-68 applied fresh, confirmed permission row + KDCA seed intact): authenticated create/update/delete round-trip on a gate, plus create on a ramp area and a taxiway, all via real HTTP requests; confirmed `editable: true` for a user holding the new permission and a 401 + `editable: false` for one who doesn't.
- Manual browser click-through of the actual draw/drag interactions was not possible this session (no connected browser-automation tool, and the app requires VATSIM OAuth login) — verified instead by tracing the interaction logic against the FCA editor it mirrors, plus the backend-side confirmation above that every endpoint it calls behaves correctly.


Vainnor:
QA review: returned for rework, fix pushed to this branch (`ddbbee1`).

**Bug found:** `SurfaceMap.tsx`'s `handleClick` had no `phase` check before appending a vertex to the current taxiway/ramp draft. Once a shape was finished (or an existing taxiway/ramp reopened just to rename it), any stray map click that wasn't a vertex-handle grab silently appended a phantom point — corrupting the geometry on the next Save. The FCA polyline editor this code explicitly mirrors guards the equivalent case with `drawMode === "draw"`; that guard was missing here. Gates were unaffected (already guarded by their own points-length check).

**Fix:** added the missing `if (phase !== "draw") return;` guard, matching the sibling editor's pattern. Also fixed a stale doc comment on `can_edit` left over from the `events.config.update` → `flow.surface_data.update` permission rename.

**Verified:** `pnpm typecheck`, `pnpm --filter web test` (88/88, no regressions), `cargo fmt --check`, `cargo clippy -D warnings` — all clean.

Everything else in this PR held up under review: the three-in-sync permission swap (marker + catalog + migration, applied atomically across all 9 handlers) is correct and `auth::permissions::sync_tests` passes; full backend suite green. Two non-blocking notes for awareness, not fixed here: the empty-airport `editable` fallback uses a non-facility-scoped permission check, but that's a pre-existing pattern already shipped in `airport-configs.tsx`, not new to this PR.

Vainnor:
**Review findings — fixed:**

Independent `/code-review` pass on top of my own manual review found real bugs in the draw/edit state machine, beyond the doc-comment + phase-check fix another concurrent session pushed to this branch in parallel (rebased cleanly, removed one resulting redundant/dead check):

1. **Vertex dedupe threshold too large.** `addVertex`'s jitter-dedupe used `haversine(last, p) < 0.05` (0.05nm ≈ 92m/304ft) with no time component — silently swallowed legitimate airport-scale geometry (gates, tight taxiway curves) closer together than that, with zero feedback that the click did nothing. Reduced to 0.001nm (~1.85m), enough to only catch a genuine same-spot double registration.
2. **Multi-ring ramp areas silently truncated.** `startEditExisting`/`save()` only ever read/wrote `rings[0]`; opening a ramp/apron area with additional rings (e.g. a hole) and saving — even just to rename it — permanently deleted every ring beyond the first. Not reachable via today's KDCA seed data or this editor's own create path, but a live data-loss trap the moment such a row exists another way. The draft now carries any extra rings through untouched and resends them on save.
3. **Migration 0068 didn't backfill existing grants.** It only inserted the new `flow.surface_data.update` permission row — anyone already holding `events.config.update` specifically to manage surface data under #177's placeholder scheme would silently lose write access the moment this ships. Backfilled `role_permissions`/`user_permissions` additively (not a rename — `events.config.update` keeps its own unrelated meaning for `airport_configs.rs`, so this grants the new permission *alongside* it), preserving each `user_permissions` row's `granted`/`artcc_id` exactly (a facility-scoped denial stays a denial, not a blanket grant).

**Added tests** (this code had none before): `layers.test.ts` for `buildSurfaceLayers`'s selection-filtering and `buildSurfaceDraftLayers`'s polygon-closing logic (mirrors the existing `fca.test.ts` pattern), and a handler test proving the permission-backfill query logic preserves `granted`/`artcc_id`. All new/changed logic mutation-tested (reverted each fix, confirmed the relevant test/assertion goes red, restored).

Also confirmed via mutation testing that the ARTCC-scope authorization check (`require_edit`) itself has zero handler-level test coverage — but that's a pre-existing gap shared verbatim by 3 handlers (`airport_configs.rs`, `airport_surface.rs`, `facility_map.rs`) predating this PR, not a #178 regression. Flagging for a separate follow-up issue rather than blocking here.

**Verified:** `cargo test --workspace` (187 backend + 17 discord + 3 core), `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `pnpm typecheck` all clean. `pnpm --filter web test`: 97/97 passing (88 → 97, 9 new tests).

Pushed to `feat/178/surface-editor`.

Vainnor:
**Third-pass findings — fixed:**

1. **Doc/code contradiction.** `AirportRampAreaBody`'s doc comment said `rings` is "the outer boundary only (holes aren't modeled)" — directly contradicted by the previous round's `extraRings` preservation fix, since nothing in `validate_ramp_area` actually rejects `rings.len() > 1`. Reworded the comment to state what's actually true: every row created so far has exactly one ring, the editor only draws/edits the first, and any further rings are preserved rather than assumed impossible.
2. **Stale double-click timer.** `lastClickT` (the 300ms double-click-finish detector) was never reset between drafts. Finishing shape A via double-click then immediately starting shape B and clicking the map could make B's first click look like a double-click too, silently no-op'ing it (`finishDraw()` on an empty draft does nothing). Narrow window in practice, but a one-line fix: reset in `startNew`/`startEditExisting`.

Everything else this pass surfaced (duplication vs the FCA editor, unmemoized layer rebuilds, 9 near-duplicate CRUD hooks, the ring-closure distance heuristic) was either a restatement of first-pass findings already judged to overlap with #186 (map rendering pipeline cleanup, already queued), or a pattern I'd already evaluated and matched to an existing accepted convention (`airport-configs.tsx`'s identical `editable` fallback) — not re-litigating those here.

**Verified:** `cargo test --workspace` (187 backend + 17 discord + 3 core), `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `pnpm typecheck` all clean, `pnpm --filter web test` 97/97.

This closes out the review — three rounds is the line for this issue; further code-quality polish is tracked separately (#186, #198). Pushed to `feat/178/surface-editor`.

Issue #164, opened 2026-09-12
Data-driven departure timing: replace the flat 8-min taxi/pushback allowance with learned per-gate/type/runway estimates


Vainnor:
Epic: replace the single fixed "8 minutes to push, start, and taxi out" assumption with **data-driven,
per-gate / per-type / per-runway** pushback + taxi-out estimates learned from history — VATSIM-outlier
robust, explainable in debug mode, browsable by staff, and resting on **editable, permissioned airport
surface data**. This is **L/XL — split into the sub-issues (A–G) below** at triage (the board supports
Parent issue / Sub-issues progress). Filed to **Blocked**: sub-issue **A** (the surface-data model) is
the foundation everything else keys on, and several sub-issues carry decisions to make at triage, so
this can't start as one unit.

## What we do today (the thing to replace)

Every not-yet-airborne flight is given a **flat 8-minute** allowance for *pushback + engine start + taxi
out*, applied uniformly regardless of aircraft type, gate, airport, or departure runway:

```rust
// backend/src/feed/predict.rs:28
/// Taxi + spool-up allowance added to a not-yet-airborne aircraft's flight time…
pub const GROUND_TAXI_SEC: f64 = 8.0 * 60.0;
```

It is added in the shared prediction service at `predict.rs:76` (so it feeds **every** ETA surface —
arrival ladder, FCA metering, runway ETE) and again in airport-flow demand (`feed/flow.rs`, the
`ft_min` term). So a heavy off a far cargo ramp at ORD and a regional off a close gate at a small field
get the *same* 8 minutes, which skews release/CFR times and demand curves.

We already have two ingredients this epic builds on:

- **A live taxi-out observer** — `backend/src/feed/taxi.rs` watches each departure from first roll (~7 kt)
  to airborne and records the elapsed time, aggregated **per airport** (rolling 3 h, in-memory) for the
  Taxi Monitor page. It does *not* key by gate / type / runway, does *not* capture the **gate-sit**
  (pushback + start, *before* movement), and isn't persisted for long-term learning. Its outlier handling
  is a crude 3 s–60 min clamp.
- **A persisted position stream** — the stats collector stores every pilot tick (lat/lon/gs/alt/heading +
  aircraft type, `reconstruct.rs::PilotRow`), which is exactly the surface-track history a learning model
  needs to derive per-gate/per-runway distributions offline.

## The model to build (FAA analog)

[...]

PR #197, opened 2026-09-13
feat(stats): persist per-gate/type/runway pushback + taxi-out observations

### LINKED ISSUE #179: Persist per-gate/type/runway pushback + taxi-out observations from the departure stream
## What happens

There's no persisted, per-(gate, aircraft type, runway) record of how long pushback+startup and
taxi-out actually take. The existing taxi observer (`backend/src/feed/taxi.rs`) only aggregates
per-airport, in-memory, over a rolling 3h window — nothing is kept long-term, and it doesn't capture
the gate-sit (pushback+start) phase at all, only movement-to-airborne.

## Why

`feed/taxi.rs` watches each departure from first roll (~7 kt) to airborne, but that starting point
already skips the pushback/engine-start phase entirely, and its output isn't matched to a specific
gate or persisted anywhere a learning model could use it later.

## What should happen

Extend the observer (or add an offline job over the already-persisted position stream —
`reconstruct.rs::PilotRow` has every tick with lat/lon/gs/alt/heading + aircraft type) to record,
per departure:
- Matched **gate** — nearest defined gate/parking spot (from sub-issue A) at spawn/first-seen.
- **Aircraft type/class**.
- **Departure runway**.
- **Pushback+startup** duration (gate-sit, before first movement).
- **Taxi-out** duration (movement → airborne, matching the existing observer's own boundary).

Persist keyed by `(airport, gate, type, runway)` so estimates can accrue over time (sub-issue D
consumes this).

## Acceptance

- [ ] Each departure's pushback+startup and taxi-out durations are persisted, matched to a gate
      (when one is defined for the spawn point), aircraft type, and departure runway.
- [ ] A departure with no matched gate (undefined surface data at that airport) still records
      taxi-out at the airport level, just without a gate key — no data is silently dropped.
- [ ] A test covers the gate-matching logic against sample spawn coordinates.

Blast radius: new persisted table (additive) + feed subsystem (`feed/taxi.rs` or a new offline job).
No existing endpoint/schema changed, no trajectory/ETA model impact yet (sub-issue E wires estimates
in).
Part of the #164 epic (sub-issue C). Depends on sub-issue A (needs gate/runway data to match against).

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Adds a new live, DB-connected background collector `backend/src/feed/taxi_observations.rs`, mirroring `feed/delays.rs`'s departure-taxi timing exactly (same thresholds: GS_START=7, GS_STOP=60, ALT_CLIMB_FT=100, DEP_PROX_NM=15), extended to also time the pushback+startup phase (first-seen → start-of-taxi) and match the spawn point to the nearest defined gate (`flow.airport_gate`, sub-issue A / #177).
- New table `stats.taxi_observation` (migration `0068`), mirroring `stats.flight_leg`'s shape exactly — nullable `gate_id`/`aircraft`/`runway`/`pushback_sec` so a departure with no matched gate, or one first observed already rolling, still records what it can rather than dropping the observation.
- Pruned by the existing `stats_compaction_once` job alongside `stats.flight_leg`, reusing the same retention cutoff — no new job, no new retention constant.
- **No new permission, no new endpoint** — matches this issue's own AC scope, and mirrors how `stats.flight_leg` itself shipped (persistence-only commit; a later, separate issue can add a read endpoint reusing the existing `stats.read` permission, same as `flight_leg`'s own history).

**Architecture note:** I initially planned an offline job over historical position data to avoid giving the feed subsystem a DB handle, based on the "feed subsystem has no DB handle" framing in CLAUDE.md. That framing doesn't hold universally — `feed/delays.rs` is already a live, DB-connected collector doing near-identical work. This mirrors that existing, already-shipped pattern instead.

Closes #179

## Test plan
- New unit tests in `feed/taxi_observations.rs` (4): pushback+taxi-out timing end-to-end, no-pushback-figure when first observed already rolling (this caught a real state-machine bug during review — the session-creation branch never checked for already-past-threshold groundspeed, which I fixed), and `nearest_gate` matching/no-match-when-too-far (AC 3's gate-matching test).
- `feed::delays` tests (3) still pass unaffected (only change there is making `nearest_runway` `pub(crate)` so this module can reuse it).
- Live check against an isolated throwaway Postgres: all migrations (1-68) applied fresh, `stats.taxi_observation` schema + FK to `flow.airport_gate` confirmed correct, backend ran cleanly with the new collector spawned (no panics/errors) for ~35s of ticks with no live feed data available in this sandbox.
- Exercising the collector against real live VATSIM traffic wasn't possible in this sandbox — verified via the unit tests (which exercise the exact same `process()` function the live collector calls) instead.


Vainnor:
QA review: returned for rework, fix pushed to this branch (`6dc2fa6`).

**This module's own doc comment claimed to mirror `feed/delays.rs`'s departure detection "exactly (same thresholds)" — it wasn't quite exact.** Two real gaps found via a direct line-by-line diff against `delays.rs`:

1. **No `arriving_turnaround` guard.** A pilot near both their filed departure and arrival airport at low speed (pattern work, touch-and-goes, short dep≈arr hops) was mistaken for a genuine pushback, injecting garbage timings into the exact training data this feature exists to produce for the future per-gate/type/runway estimator (#164 sub-issue D). `delays.rs` already guards against this; ported the same check.
2. **No short-roll fallback.** When Watching→Rolling and the stop condition land in the same 15s poll tick, `start_ms == now_ms` gives `dur == 0`, which silently fails the `MIN_TAXI_SEC` bound and drops the whole observation — a systematic bias against fast-accelerating departures. `delays.rs` already falls back to `first_seen_ms` for this case; ported the same fallback.

Also fixed the collector's inline `airport_surface::list_gates` DB read, which directly violated AGENTS.md's "feed functions have no DB handle... never queried inline" invariant (the file already correctly cached `airports`/`runways` this way, just not gates). Replaced it with a proper `AppState`-level `ArcSwap<HashMap<icao, Vec<AirportGateBody>>>` cache, a new `jobs::spawn_airport_gates_refresh` job mirroring the existing `aircraft_profiles` cache pattern, and force-reload-on-write in `handlers::airport_surface`'s three gate mutation handlers.

**Verified:** added 2 new tests for the fixed scenarios (`arriving_turnaround_does_not_start_a_departure_session`, `very_short_roll_in_one_tick_falls_back_to_first_seen_instead_of_dropping`) — both pass, and all 6 `taxi_observations` tests + all `delays` tests pass unaffected. `cargo fmt --check`, `cargo clippy -D warnings`, full backend test suite all clean. Live-booted against two separate isolated Postgres databases (one to reproduce the original migration-68 conflict from a concurrent worktree, one fresh) — migrations apply cleanly, the collector and the new gate-refresh job both run without error.

Everything else in this PR held up well: the additive migration, `insert_taxi_observations`'s bound parameters, the pruning-job integration, and the original "already rolling on first seen" fix (which I separately mutation-tested by reverting it and watching the test go red) were all correct as shipped.

One coordination note carried over from the original PR body: this branch and `feat/178/surface-editor` independently used migration number `0068` off the same #177 base. #178 is currently back in rework too — whichever of the two merges into `next` second will need to renumber its migration.

PR #199, opened 2026-09-13
fix(integrations): scope Discord channel_id/role_id resolution by facility

### LINKED ISSUE #194: Discord channel_id/role_id resolve a logical name across ALL guilds, not per-guild — a 2nd guild reusing a generic name silently loses its mapping
## What happens

`integration_repo::channel_id(pool, name)` and `role_id(pool, name)`
(`backend/src/repos/integration.rs:128-151`) resolve a logical channel/role name **across every
configured guild**, not scoped to any particular guild:

```rust
"select ch.channel_id from integration.discord_channels ch \
 join integration.discord_configs c on c.id = ch.config_id \
 where ch.name = $1 order by c.created_at limit 1"
```

If two different guilds each define a channel or role under the same generic logical name (e.g.
both configure an `aceteam-requests` channel, or an `ntmo` role), only the **first-created**
guild's mapping is ever returned — the second guild's row for that name is silently unreachable.
No error, warning, or log line surfaces this anywhere.

## Why

`integration.discord_channels`/`discord_roles` enforce `unique(config_id, name)`
(`backend/migrations/0048_integration_discord.sql`), not a global unique on `name` — the schema
explicitly permits two guilds to define the same logical name. But every caller
(`handlers/ace.rs`, `handlers/tmu.rs`, `handlers/events.rs`) calls `channel_id`/`role_id` with just
the bare name, never a guild/facility hint, so the `order by created_at limit 1` in the query is
the only tiebreaker — an accident of insertion order, not a real routing decision.

The migration's own header even calls this out: `-- discord config (one guild for the first cut,
but modeled for many)`. This is that gap, still open.

## Why the neighbouring path is fine

The per-region event-thread channel (`events.rs:1497-1504`, `region-{region}` names, falling back
to the generic `events` name) already sidesteps this by giving each destination its own unique
logical name rather than relying on guild-scoping. That pattern works *because* the names it uses
happen not to collide across guilds today — it isn't evidence the underlying resolution is
guild-aware, just that this one call site avoided the trap.

## What should happen

Thread an explicit scope (most naturally the same `artcc`/facility scoping already used for
permission checks elsewhere) through `channel_id`/`role_id`, so a lookup can prefer the guild whose
config maps to the relevant facility and only fall back to the generic name when no facility-scoped
guild is configured — mirroring how `events.rs`'s region routing already falls back to `events`.
Alternatively (simpler, more restrictive): make `name` globally unique across
`discord_channels`/`discord_roles` so a collision fails loudly at config-write time instead of
silently at read time — cheaper to build, but blocks legitimately wanting two guilds to each have
their own e.g. `events` channel.

Found during the #151 audit (verifying multi-guild routing has no hard-coded ids) — no hard-coded
snowflake exists anywhere in `ois-discord`/backend, but this ordering-dependent collision is a
real, related routing gap the audit surfaced.

## Acceptance

- [ ] A second guild configuring `aceteam-requests` (or any other currently-generic name) routes
      correctly for its own facility's requests, instead of silently losing to the first-configured
      guild.
- [ ] A test reproduces the current bug (two guilds, same logical name, second one unreachable)
      and passes once fixed.
- [ ] `docs/features/discord-integration.md`'s "Config & account linking" section is updated to
      match whatever routing model is chosen here.

Blast radius: `backend/src/repos/integration.rs`'s `channel_id`/`role_id` + every caller
(`handlers/ace.rs`, `handlers/tmu.rs`, `handlers/events.rs`) — no API contract change if the fix
stays backend-internal (adding an optional facility param to a repo function). No
permission/three-in-sync impact. No trajectory/ETA impact.
Pre-existing; found while auditing #151.
Relates to #143, #151.

Vainnor:
## Summary
- `integration.discord_configs` had no facility association, so `channel_id`/`role_id` (`backend/src/repos/integration.rs`) could only tie-break same-named channels/roles across guilds by `created_at` order — a second guild configuring an already-used name (e.g. `aceteam-requests`) was silently unreachable regardless of which facility it actually served.
- Adds `integration.discord_config_facilities` (migration `0068`, many-to-many guild↔ARTCC) and an optional `facility` param on `channel_id`/`role_id`: a facility match wins over the `created_at` fallback, so the guild serving that facility is preferred instead of always losing to whichever guild was configured first. `facility: None` (or one no guild claims) falls back to exactly today's behavior — fully backward compatible.
- Threaded the new param through call sites with an unambiguous facility already in scope (ACE requests scoped by the target/neighbor ARTCC, event-thread/TMU-package publish scoped by the host); left two call sites unscoped with a documented reason — the generic `ntmo`/`dcc-trainee` event-thread staff roles, and the TMU-advisory channel (a TMI has `requesting`/`providing` ARTCCs but no single issuing facility).
- Extends the Discord config admin page (`web/src/pages/admin/discord.tsx`) with a per-guild facility picker; regenerated the API client for the new `facilities` field.
- `docs/features/discord-integration.md` updated to describe the new table and the fallback behavior.

**Note for whoever merges after this**: sibling branches `feat/178/surface-editor` and `feat/179/taxi-observations` also use migration `0068` off an earlier base — whichever of the three merges last will need to renumber during rebase.

Closes #194

## Test plan
- New `#[sqlx::test]` in `repos/integration.rs` reproduces the bug (two guilds, same channel name, second one unreachable) and confirms the fix: a facility-scoped lookup returns the matching guild, while an unscoped or non-matching lookup still falls back to the first-configured guild.
- Full backend test suite: 192 passed, 0 failed (includes `tests::migrations_apply_cleanly`, confirming the new migration applies cleanly in the full chain).
- Client regenerated against a running backend (isolated throwaway Postgres); `pnpm --filter web typecheck` and the full web test suite (104 tests) pass.
- Manual click-through of the new admin UI facility picker wasn't possible this session (no browser tool, VATSIM OAuth required) — verified via typecheck + tracing the state wiring, which mirrors an existing sibling pattern (`airport-configs.tsx`'s ARTCC list).


Vainnor:
**Review findings — fixed:**

Independent `/code-review` pass found three real gaps, plus I resolved a migration-numbering collision myself:

1. **Migration renumbered 0068→0069.** #178's branch (already approved, in Testing Queue) independently claimed migration 0068 off the same base (`next` currently ends at 0067). Renumbered this one to avoid a collision when both eventually merge.
2. **Unnormalized facility input.** `event.facility` is a raw passthrough of the external VATUSA v3 events API with no case/format guarantee — `dcc_region()` already defensively uppercases the same field for exactly this reason — but `discord_config_facilities.artcc_id` is stored uppercase and matched exactly. Passing the raw value into the new facility-scoped `channel_id`/`role_id` calls would silently fail to match on any casing mismatch, quietly falling back to the pre-#194 behavior with no error — undermining the very fix this PR makes. Now normalized via the existing `normalize_facility()` helper (made `pub(crate)`, matching how `normalize_icao`/`owning_artcc` are already reused cross-module) at all three call sites.
3. **Looser facility validation in `upsert_config`.** The new per-guild facility-insert loop validated ARTCCs more loosely (trim+uppercase+non-empty) than `normalize_facility()` (also caps length, rejects non-alphanumeric) — could silently store junk that can never match. Now reuses the same helper.
4. **`channel_id`/`role_id` duplication + untested `role_id` facility path.** The two functions had near-identical hand-copied SQL — a future edit to one's ordering logic could silently miss the other, and `role_id`'s facility-scoped branch had zero test coverage (only `channel_id` was tested, and nothing in the codebase currently passes `Some(facility)` to `role_id`). Extracted a shared `resolve_scoped_id` helper and added a `role_id`-specific test proving its own table/column wiring (mutation-tested: swapping in the wrong table name fails the new test with a real DB error).

Two other findings (an intentional, already-documented asymmetry between `activate_package` and `publish_tmi`'s scoping, and a bounded N+1 query pattern in `generate_tier1`) were judged as accepted trade-offs, not bugs — not acting on those.

**Verified:** `cargo test --workspace` (193 backend, up from 192 — 1 new test), `cargo fmt --check`, `cargo clippy --all-targets -D warnings`, `pnpm typecheck`, `pnpm --filter web test` (104) all clean.

Pushed to `fix/194/guild-scope`.

PR #201, opened 2026-09-13
feat(stats): robust per-gate/type/runway taxi estimator

### LINKED ISSUE #180: Robust per-gate/type/runway pushback+taxi estimator with outlier rejection and fallback ladder
## What happens

Sub-issue C's raw observations need to become usable point estimates, robust to VATSIM-specific
noise (pilots going AFK on the ramp, warping, spawning mid-taxi, holding indefinitely) — a naive
average would let a handful of bad samples dominate a gate's number.

## Why

VATSIM has no equivalent of FAA surface surveillance's data quality guarantees; the existing taxi
observer's own outlier handling is described as "a crude 3s–60min clamp," not a real robust
estimator.

## What should happen

From C's samples, compute per-(gate, type, runway) pushback and taxi-out estimates using a robust
statistic (median, trimmed mean, or MAD-based rejection — pick one at implementation time and
justify it) instead of a naive mean, with sane absolute bounds as a backstop.

Define the **fallback ladder** used when a specific key has too few samples to trust:
`(gate, type, runway) → (airport, runway) → (airport) → today's flat 8-minute default`.

## Acceptance

- [ ] A pure estimator function takes a sample set and returns a robust estimate, with a documented
      minimum-sample threshold below which it falls back to the next ladder tier.
- [ ] The fallback ladder is implemented exactly as `(gate,type,runway) → (airport,runway) →
      (airport) → 8min default`.
- [ ] A regression test pins both the outlier-rejection behavior (a few injected extreme samples
      don't swing the estimate) and the fallback-ladder ordering (a naive-mean or missing-guard
      revert turns it red).

Blast radius: new pure estimation logic + repo functions reading sub-issue C's data. No trajectory/
ETA model wiring yet (sub-issue E) — this issue only produces the numbers, doesn't consume them.
Part of the #164 epic (sub-issue D). Depends on sub-issue C.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Adds a pure estimator `backend/src/feed/taxi_estimate.rs` that turns sub-issue C's raw pushback/taxi-out observations (`stats.taxi_observation`) into point estimates.
- Robust statistic: **median** — a 50% breakdown point (up to half the samples in a bucket can be arbitrary outliers without moving the estimate), and it needs no tunable parameter the way a trimmed mean or MAD-rejection threshold would.
- Fallback ladder exactly as specified: `(gate,type,runway) → (airport,runway) → (airport) → 8min default`, with a documented `MIN_SAMPLES = 5` threshold per tier. Pushback and taxi walk the ladder **independently** since not every observation carries a pushback figure.
- An absolute bounds clamp backstops each tier's median (1-30 min taxi, 0-20 min pushback) against a whole bucket being systematically bad, on top of (not instead of) the median's own outlier resistance.
- `repos::stats::taxi_samples_for_airport` fetches the sample set. No consumer wiring (sub-issue E), no new endpoint, no new permission.

**Stacked on #197** (sub-issue C) — based here off `feat/179/taxi-observations` so the diff stays scoped to this issue; please retarget to `next` once #197 merges.

Closes #180

## Test plan
- 5 new unit tests (no DB — pure function): outlier rejection (verified the test actually distinguishes median from mean: mean of the injected sample set is 366s, median is 90.5s, assertion window is 85-95s), most-specific-tier-wins when every tier qualifies, exactly-one-tier fallback on sparse data, default-to-flat-8-min when every tier is sparse, and pushback/taxi walking the ladder independently.
- Full backend test suite: 197 passed, 0 failed (includes `tests::migrations_apply_cleanly`).


Vainnor:
QA review: rebased onto the current \`feat/179/taxi-observations\` (this PR's own commit is unchanged, `850910a`).

**Found a real, reproduced build breakage:** this branch was stacked on an older commit of #197's branch, from before that PR's migration got renumbered twice (0068 → 0069, on collision with #178; then 0069 → 0070, on collision with #194 — both merged into `next` while this PR sat open). Merging as-is would have reintroduced the old `0068_stats_taxi_observation.sql`, colliding with `0068_flow_surface_data_permission.sql` and breaking `migrations_apply_cleanly` — reproduced this exact failure before fixing it.

**Fix:** rebased onto the current `feat/179/taxi-observations` tip. This PR's own diff never touched migrations (only `feed/mod.rs`, `feed/taxi_estimate.rs`, `repos/stats.rs`), so the rebase was clean with zero conflicts — the stale migration file was only ever present via inherited branch history, not anything this PR added.

**Reviewed the actual estimator logic, unaffected by the rebase:**
- Verified `GROUND_TAXI_SEC` reuses the existing constant from `predict.rs` (480s = 8 min) rather than duplicating it, matching AC2 exactly.
- **Mutation-tested `median` → naive mean** myself: 3 of 5 tests fail, and the mean value my mutation produced (366.25s) matches your own stated figure (366s) almost exactly — confirms the outlier-rejection test is genuinely load-bearing per AC3, not decoration.
- Fallback ladder correctly short-circuits each tier when its key component is `None` (e.g. no gate match skips straight past the gate/type/runway tier), and the `MIN_SAMPLES` threshold is applied uniformly per-tier.
- `taxi_samples_for_airport`'s query is properly parameter-bound.
- Full suite: 205/205 passing, `cargo fmt --check`/`clippy -D warnings` clean, `migrations_apply_cleanly` passes.

No design concerns with the estimator itself — the median choice and ladder structure are sound and well-justified. The only issue was the stale base, now fixed.

PR #202, opened 2026-09-13
fix(integrations): harden #194's facility scoping (FK, normalization, validation)

Vainnor:
## What happened

#194/#199 (Discord `channel_id`/`role_id` facility-scoped resolution) merged to `next` while a second `/code-review` pass on the PR was still in progress. This PR carries the fixes from that pass that never made it into #199.

## Changes

- **Missing FK.** `integration.discord_config_facilities.artcc_id` had no foreign key to `org.facilities`, unlike every sibling ARTCC-id column in this schema (`tmu.tmis.artcc_id`, `facility_documents`, ...). A typo'd or deleted ARTCC could be stored and would silently never match any real event/request facility — reproducing the exact "silently loses to another guild" failure #194 exists to fix, just via a data-entry mistake. New additive migration (`0070`, since `0069` already merged) adds the constraint retroactively; the table is brand new so no existing-row cleanup was needed.
- **Missed normalization in `ace.rs`.** `create_request` and `enqueue_notify` pass an unnormalized ARTCC (trim+uppercase only) into the facility-scoped `channel_id` calls — the same gap already fixed in `events.rs`'s three call sites during the first rework round, just missed here. Now normalizes only the routing hint via the existing `normalize_facility()` helper; the stored/displayed `artcc_id` value's format is unchanged.
- **Silent validation drop.** `put_discord_config` silently dropped an invalid facility string instead of rejecting the request — a caller got a 200 with no signal anything was dropped. Now rejects with 400 up front, mirroring the handler's existing name/guild_id validation.
- Updated the two facility-routing tests (merged in #199) to use real seeded ARTCC codes instead of placeholder strings (`'ZZZ'`/`'XYZ'`), since the new FK now rejects those.

## Verification

- `cargo test --workspace`: 194 backend tests green.
- `cargo fmt --check`, `cargo clippy --all-targets -D warnings`: clean.
- `pnpm typecheck`: clean (no web changes in this PR).
- FK constraint's rejection behavior confirmed via a manual throwaway test before this commit (removed after confirming).

Blast radius: `integration.rs` (handlers + repo), `ace.rs` — no API contract change, no new permission, no trajectory/ETA impact.

Relates to #194.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EmJvaKhNVEK8cUCAVJsgb9


Vainnor:
**Second-pass findings — fixed:**

1. **Format-valid but nonexistent facility → opaque 500, not the promised 400.** `put_discord_config`'s new rejection only checked `normalize_facility`'s shape (length/alphanumeric), not whether the ARTCC actually exists — a well-formed but typo'd code (e.g. \"ZDX\") passed that check, then hit the new FK constraint from this same PR and surfaced as a generic 500. Now also checks existence via `org_repo::find_facility`.
2. **Inaccurate migration comment.** Claimed "every sibling ARTCC-id column in this schema" has this FK — false (`ace.requests.artcc_id`, `access.api_key_permissions.artcc_id` are deliberately free text). Corrected to state the real, narrower justification: this column is part of a composite PK in a pure join table.

**Verified:** full backend suite (194 tests), fmt/clippy clean.

Two lower-priority code-quality observations from this pass — not acting on them, noting for visibility:
- `normalize_facility` must be manually re-applied by every `channel_id`/`role_id` caller rather than being enforced once inside the shared resolver (this PR's own ace.rs fix is evidence the pattern is fragile) — a real design smell, but restructuring it is a bigger change than this hardening pass warrants.
- The `let facility_hint = artcc.as_deref().and_then(normalize_facility);` line is duplicated verbatim at both `ace.rs` call sites — minor, could be a tiny shared helper.
- `on delete cascade` on the new FK means deleting/renaming a facility silently drops its Discord routing scope — flagged by the reviewer itself as likely-intentional/low-severity, matching this schema's existing cascade convention.

Pushed to `fix/194/facility-hardening`.

PR #207, opened 2026-09-13
🔴 Hotfix: align react to 19.3.0 (blank page — React #527 react/react-dom mismatch)

Vainnor:
## 🔴 Production hotfix — ois.vzdc.org is a blank page

After release #205, the site loads to a **blank white page for every user**. Console shows a single
fatal, app-crashing error:

```
Uncaught Error: Minified React error #527
```

Decoded, #527 is: **"Incompatible React versions: the `react` and `react-dom` packages must have the
exact same version."**

## Root cause

Dependabot #169 bumped **`react-dom` to 19.3.0** but left **`react` at 19.2.8** (in both
`web/package.json` and `packages/ui/package.json`). React **19.3** added a runtime exact-match check
between `react` and `react-dom`; with the two a minor apart it throws on mount and, with no root error
boundary, blanks the whole app. The skew rode into `main` with release #205.

(Flagged at the time #169 merged: "react-dom to 19.3.0 while react stays 19.2.8 … worth landing a
matching react 19.3.0 to keep them in lockstep." This is that follow-up, now as a hotfix.)

## Fix

- `web/package.json` and `packages/ui/package.json`: `react` and `@types/react` → `^19.3.0` (matching
  the already-present `react-dom` `^19.3.0`).
- `pnpm-lock.yaml` refreshed — **every** `react`/`react-dom` now resolves to `19.3.0(react@19.3.0)`.

## Verification

- `pnpm --filter web typecheck` — pass.
- `pnpm --filter web build` — pass.
- All react/react-dom entries in the lock resolve to 19.3.0; the version-mismatch invariant (#527) can
  no longer fire.

## Acceptance

- [ ] ois.vzdc.org renders again (no React #527) after this deploys.
- [ ] `react` and `react-dom` resolve to the same version in `pnpm-lock.yaml`.

Blast radius: web dependency versions only (`web/`, `packages/ui/`, lockfile). No source/behavior
change, no backend, no API contract, no migration.
Hotfix straight to `main` because production is down. **Follow-up: the identical skew is still on
`next`** — it must get the same bump (or cherry-pick this) so the next release doesn't reintroduce the
outage.


Vainnor:
Superseded by #208 (same fix routed through `next`), which will then be promoted via a `next → main` release PR per the normal flow. Closing this direct-to-main hotfix.

PR #236, opened 2026-09-14
fix(flow): landed arrivals no longer re-appear in their destination FCA

### LINKED ISSUE #213: Landed arrivals still show in their destination FCA — ground aircraft use the full route (JBU254 at KJFK re-appears in CAMRN)
## What happens

An aircraft that has already **landed at its destination** still appears in that airport's FCA as
crossing traffic. Observed: **JBU254** (KDCA→KJFK) is on the ground at KJFK (FL0, 5 kt, past CAMRN) yet
shows under the **JFK [CAMRN#]** FCA as a GND crossing. An aircraft that has already passed through an
FCA must not reappear in it — it's done.

## Why

FCA membership is a geometric crossing of the aircraft's resolved route against the FCA line
(`backend/src/feed/fca.rs`: `crossing_for` → `route_path` → `crosses`/`path_crossing`). For an
**airborne** aircraft the route is trimmed to the portion still ahead, so a past crossing drops off. But
for a **ground** aircraft the *full* filed route is used, with no forward-trim:

```rust
// backend/src/feed/fca.rs:168 — route_path
let path = if gs >= 50 {
    remaining_anchors(&anchors, lat, lon, hdg as f64)  // airborne: remaining route only
} else {
    anchors                                            // ground: FULL dep→arr route
};
```

The `gs >= 50` branch assumes a ground aircraft is *pre-departure at the origin* (a legitimate future
crossing). It doesn't distinguish that from an aircraft that has **arrived and is on the ground at the
destination** — which still carries the whole KDCA→…→CAMRN→KJFK route, so `path_crossing` still finds
the CAMRN crossing and the flight re-appears in the FCA from the far side. (The heading gate the code
avoids on purpose — see the `crosses` comment at `fca.rs:178` — isn't the issue; the route *is* used.
The gap is specifically the un-trimmed ground route.)

## What should happen

Only count an aircraft against an FCA if the crossing is still **ahead of it on its route**, for ground
aircraft too — i.e. project the aircraft's current position onto its filed route and keep only the
forward remainder before testing the crossing, so a flight that has already passed the FCA point (or
landed at the destination) yields no crossing.

Trade-off to weigh (why ground isn't already trimmed): `remaining_anchors` trims using **heading**,
which is unreliable for a parked/slow aircraft — that's why the `gs>=50` gate exists. Options:

- Trim by **along-route position** (nearest point on the polyline to the current lat/lon), not heading —
  robust at any speed, and the natural "project forward on the route" the report asks for. A pre-departure
  aircraft at the origin projects to the route start → full remainder → still crosses (correct); a landed
  aircraft projects to the destination end → nothing ahead → no crossing (fixed).
- Or a narrower guard: exclude a ground aircraft whose position is within ~N nm of its **destination**
  (arrived), which kills the JBU254 case without touching pre-departure grounds. (Mirrors the
  arrived-at-destination guard already in `feed/taxi.rs`.)

The along-route-projection approach is the more general fix and matches the intent.

## Acceptance

- [ ] A flight on the ground at its destination (past the FCA) does **not** appear in that FCA.
- [ ] A pre-departure flight still on the ground at its origin **does** appear (future crossing kept).
- [ ] An airborne flight that has already crossed the FCA still drops off (no regression).
- [ ] A unit test in `fca.rs` covers "landed at destination on a route that geometrically crosses the FCA
      → no crossing" (turns red if the ground branch reverts to the full route).

Blast radius: **the trajectory/FCA crossing model** — `backend/src/feed/fca.rs` `route_path`/`crosses`
is resolved once per aircraft and tested against every FCA, and feeds metering/CFR sequencing; verify
metered demand and the ladder still look right after the change. No API contract change.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- Ground aircraft's route was never trimmed for FCA-crossing purposes, so a flight that landed at its destination still carried its whole filed dep→arr route — a flight past an FCA (e.g. JBU254, landed at KJFK, still showing under the JFK CAMRN FCA) reappeared as if still approaching it.
- Adds `project_forward_index`/`forward_route_from_position` to `backend/src/feed/fca.rs`: trims a ground aircraft's route by along-route position only (never heading, which is unreliable for a parked/taxiing aircraft). A pre-departure aircraft keeps the whole route ahead; a landed aircraft's path collapses to nothing, so it can't match a crossing behind it.
- The first pass of this fix used a coordinate-epsilon check to avoid a side effect (see below) — review caught that it only worked for the exact ARP coordinate, not a realistic gate position, so it's replaced with a structural, index-based rule instead.

## Test plan
- New unit tests in `fca.rs`: landed-at-destination drops the crossing; pre-departure-at-origin still matches it.
- Caught and fixed a regression against a sibling trajectory-model caller during review: the naive first pass silently dropped `predict::arrival_eta`'s ground-route padding for an unresolved route once the aircraft sat at (or, in the corrected fix, anywhere near) an endpoint anchor. Added a realistic ~0.4nm gate-offset regression test in `predict.rs` proving the fix holds for real positions, not just an exact coordinate.
- Verified all three `feed/trajectory.rs`-adjacent `route_path` callers per `AGENTS.md`'s invariant: `fca.rs`'s own suite (17 tests), `predict::arrival_eta`'s suite (unchanged tests + 1 new), `feed::flow`'s suite (18 tests, unchanged, including `ground_flight_gets_wheels_up_cfr`). `runway.rs` never calls `route_path` with a ground aircraft.
- `just ci` green: fmt, clippy, 223 rust tests (against a per-worktree local postgres), pnpm typecheck.

Closes #213


Vainnor:
Two more commits after re-review (this PR was Returned):

- `bce1374` — removed a duplicate departure-airport lookup that `b5dc1cd` left behind in `build_candidates` (same value already in scope, just reused it).
- `1e88cfb` — hardened `prefile_position()`: it previously fell back to a fabricated `(0.0, 0.0)` when a prefile's departure airport didn't resolve, reasoning that `route_path` would bail via its own anchor-count check. That's only true when the *whole* route resolves to under 2 anchors — `nav::build_anchors` only skips the departure-specific anchor, so a resolvable arrival or route content could still reach `route_path`'s ground trimming with a Null Island position. `prefile_position` now returns `Option<(f64, f64)>` and both call sites (`fca_counts`, `build_candidates`) skip that prefile instead of guessing, matching `feed::flow::ground_estimate`'s actual (stricter) pattern.

Also traced every production call site of `route_path` (9) and `arrival_eta` (5) in the backend this pass to confirm nothing else shares this class of bug — all are airborne-only, use a live pilot's real position, or are now correctly gated.

`just ci` green (fmt/clippy/226 rust tests, pnpm typecheck).

PR #250, opened 2026-09-14
feat(flow): surface a departure's taxi/pushback derivation in debug mode

### LINKED ISSUE #182: Debug-mode breakdown of a departure's taxi/pushback timing derivation
## What happens

Once sub-issue E replaces the flat 8-minute allowance with a learned, multi-tier-fallback estimate,
there's no way for staff to see *how* a given departure's timing was derived — which gate/type/
runway matched, how many samples backed it, the pushback-vs-taxi split, or which fallback tier fired.

## Why

A silently-wrong or silently-thin estimate is worse than the old flat constant if nobody can tell
the difference — debug mode is this app's existing pattern for surfacing internal derivation (see
the profile/headwind block in `web/src/pages/fca/detail.tsx`).

## What should happen

Add a new `debug` user setting if one doesn't already cover this (check
`web/src/lib/settings/registry.ts` for an existing debug-mode setting from #39 first — reuse it,
don't add a second one). When debug mode is on, surface the derivation breakdown: matched
gate/type/runway, sample count backing the estimate, the pushback-vs-taxi split, and which fallback
tier was used. Backend exposes this breakdown on the prediction payload (new fields → **regenerate
the client**).

## Acceptance

- [ ] With debug mode on, a departure's predicted timing shows its derivation (matched key, sample
      count, pushback/taxi split, fallback tier) somewhere near the existing debug surfaces.
- [ ] With debug mode off, the breakdown fields exist on the payload but aren't rendered (no UI
      change for normal users).
- [ ] The typed client is regenerated after the payload change.

Blast radius: backend (`predict.rs`'s payload gains debug fields) + API contract (client regen) +
web (debug-mode UI). No trajectory/ETA *behavior* change — this only exposes what sub-issue E
already computes.
Part of the #164 epic (sub-issue F). Depends on sub-issue E.

🤖 Drafted by Claude Code

Vainnor:
## Summary
- #220 (sub-issue E) computes a learned per-(gate,type,runway) ground allowance, but `resolve_ground_allowance_sec` collapses the whole derivation — matched gate/runway, sample counts, pushback/taxi split, fallback tier — down to one `f64` before it ever leaves `feed::flow`. With debug mode on there was no way to see how a number was derived, or tell a well-supported estimate from one silently resting on the flat default.
- Splits the resolver: a new `resolve_ground_allowance()` returns a `GroundAllowanceBreakdown`; `resolve_ground_allowance_sec` becomes a thin wrapper over it, so its one production caller (`ground_estimate`) and existing tests are unchanged.
- `handlers::flow::build_candidates` threads the breakdown through to `fca_debug` (`None` for an airborne flight, which never applies one) as a new `FcaFlightDebug.taxi_estimate` field, rendered in the existing debug block in `fca/detail.tsx` alongside the profile/TAS/headwind line.
- No trajectory/ETA behavior change — this only exposes what #220 already computes.

## Test plan
- New `resolve_ground_allowance_exposes_the_matched_key_and_tier` covers both a clean gate+type+runway match and the thin-data default-tier fallback.
- Client regenerated (`FcaFlightDebug`'s schema changed) — verified the regenerated schema is **byte-identical** to what CI's `client-drift` job itself produces (`cargo test openapi::dump_openapi_json` → codegen from that dump), not just eyeballed against a live-backend regen.
- `just ci` green: fmt, clippy, 226 rust tests (ran against a per-worktree local postgres), pnpm typecheck, pnpm test, full `pnpm build`.
- No live browser was available this session, so the debug panel's rendering wasn't visually confirmed — verified via typecheck/build and tracing the render logic against the new field shapes instead.

Closes #182


Vainnor:
Superseded by #265 — this PR is unmergeable (`next`'s history was rewritten after this branch was cut; GitHub reports "no history in common" and refuses to reopen it). #265 is the same feature commit, cherry-picked cleanly onto current `next`, independently re-verified. Closing.

